PR #27, opened 2022-01-27
Example data

### LINKED ISSUE #10: Create examples of raw and annotated tabular data
Have one or two realistic examples of raw (BIDS)

For age have an example:
- [x] with BIDS-conform age (float in years)
- [x] where age is not in years (but e.g. months)
- [x] of age as a range of values
- [ ] of age as a lower bound (with a `+`)
- [x] of age as a somewhat datetime string (ISO 8601)
- [x] of a column called age that clearly is mislabeled

For gender have an example
- [ ] for each BIDS conform version (e.g. f, F, female, FEMALE, Female)
- [ ] for BIDS conform sex but numerically mapped via data dictionary
- [x] for nonsensical gender (i.e. float, clearly mislabeled values)

surchs:
Added some examples for participants.tsv data and (most of) the corresponding json sidecar files. All of the files are real data from openneuro (hence the ds dataset IDs). 

(mostly) closes #10


surchs:
@jarmoza: let me know where you were thinking these files should go.

jarmoza:
@surchs I think these files should go in the 'assets' folder at root. Maybe put them in their own 'example' subfolder? The assets folder does not yet exist in the main branch.

One issue that needs to be addressed is that the sidecare json files from https://github.com/metaneuro/annotation_tool/pull/27/commits/431aee0539f3e6234acbb723717d31be82fc12cc all just contain 404 messages.

surchs:
@jarmoza: thanks for the points. 
> I think these files should go in the 'assets' folder at root

I think that's where this PR would put them. Let me know if you think a different place is better
> One issue that needs to be addressed is that the sidecare json files from 431aee0 all just contain 404 messages.

Yes, I noticed this too. Not quite sure why, at least some of these files should exist. Will check my downloading code.

surchs:
I think two things were happening here: 
1. I ran a curl script to get participant.json files for every participant.tsv file. I didn't check whether these files actually exist so for the ones that don't, we got the 404
2. Some of the participant.json files that returned 404 actually exist at the requested address. I haven't really found out why they failed to download the first time (and why some were downloaded successfully the first time). Rerunning this got me the remaining two that exist. 

So I removed the participant.json files that don't exist, and added the ones that do. I think a lot of "real world" BIDS datasets come without the participants.json data dictionary. I am not sure that we should support these cases because it'll probably make our lives harder. But I'll leave them in for now since we also have examples with data dictionaries

jarmoza:
@surchs Looks good now. As for whether or not we want to support datasets without a participants.json file, we should decide that. For now, I already have logic built into my incoming PR that allows for just using a participants.tsv file.

jarmoza (approved):
Still getting used to the reviewing system. All looks good!

PR #54, opened 2022-02-18
move examples to their own root dir

### LINKED ISSUE #48: Example tsv and json files need to be moved out of the assets folder
Npm issues warnings about the example files in our `assets` folder when it attempts to compile the project for running. This is likely since `assets` is a folder the contents of which are parsed by Vue/Nuxt for optimization of serving them (typically external css, js, and images).

Our example files should just be moved to a sibling folder that Vue/Nuxt does not read, i.e. `examples`.

surchs:
- fixes #48


jarmoza (approved):
Tested and approved

PR #61, opened 2022-02-23
Page state preservation for Categorization page, Palette/Table clicking simplification, and general cleanup for landing and categorization pages

jarmoza:
1. Fix for preserving page state of categorization page - table is (re)painted on page load based on info in the store (see 3 below)
2. Retrieval of palette on page load (see `mounted()` in each page) from stylesheet and placement of palette in store for later use by JS components
    - Current colors for table painting are immediately set afterward (`setCurrentPaintClass(0)` in `categorization.vue`) and placed in the store
3. Replacement of old way(s) of table painting with simplified method
    - Table row is styled appropriately upon click
    - Row ID is placed in store upon table row click and easily identified and repainted (see `paintTable()` in `categorization.vue`)
4. Removal of old and now-unused functions and data
5. General streamlining of landing and categorization pages/components

### DIFF
--- a/pages/categorization.vue
+++ b/pages/categorization.vue
@@ -8,7 +8,7 @@
 
 			<b-col cols="4">
 				<coloring-listgroup
-					:categoryData="recommendedCategories"
+					:categories="recommendedCategories"
 					:defaultPalette="$store.state.pageData.categorization.default"
 					tag="recommended-column"
 					title="Recommended Categories"
@@ -21,9 +21,8 @@
 			<b-col cols="8">
 				<filedata-table
 					:fields="columnMatchTableFields"
-					:paintClass="getPaintClass"
-					:currentPalette="$store.state.pageData.categorization.current"
-					:defaultPalette="$store.state.pageData.categorization.default"
+					:currentPalette="$store.getters.currentPaintBrush"
+					:defaultPalette="$store.getters.defaultPaintBrush"
 					:tableData="tableDataFromTsvAndOrJson"
 					:tableID="tableID"
 					v-on:column-name-selected="tableClick($event)">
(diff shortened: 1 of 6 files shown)


pages/categorization.vue
```diff
@@ -350,17 +327,13 @@
...
-				let defaultForegroundColor = this.$store.state.pageData.categorization.default.fColor;
-
 				// 0. Counts number of painted columns
 				let columnCount = 0;
 
 				for ( let columnName in paintingData ) {
 
-					if ( defaultBackgroundColor != paintingData[columnName].bColor || 
-						 defaultForegroundColor != paintingData[columnName].fColor ) {
+					if ( this.defaultBackgroundColor != paintingData[columnName].bColor ||
```
surchs, line 335: I really think we need to separate the categorization (and how we store categorization information in the store) from anything related to color, styling, and CSS. At the moment, if I understand things correctly, we have a global store called `pageData.categoriztion.paintingData` that keeps track of `tsvCategory` and `fColor` and `bColor` and `primaryKey` for the rows that have been annotated. 

And here we are counting columns that have been annotated with `countPaintedColumns` by comparing their color names to a default color name. But the color is just a visual guide, the important information is what category a column has been mapped to. So I really think that routing everything through the color adds a lot of hard to understand complexity here that we should think about clearing up before we merge this!

store/index.js
```diff
@@ -116,12 +133,64 @@ export const actions = {
...
+			}
+		}
+
+		// 3. Set the new palette
+		p_context.commit("setPalette", {
+			backgroundColors: backgroundColors,
+			foregroundColors: foregroundColors
+		});
+	},	
+
```
surchs, line 193: If I understand this correctly, the entire above added section (`retrievePalette`) exists in order to regex parse the main.css stylesheet in order to retrieve the colors we have defined there and store them in a global store `backgroundColors` and `foregroundColors`. And if I also understand this correctly, the only reason we need to do that is because we keep track of what category a column has been assigned to by comparing it's current DOM style color attribute to the values in this list.

As I mentioned above, I really think we need to completely split colors and categorization and focus on keeping track of the categorization.

surchs (changes requested):
Thanks @jarmoza for the PR. I took some time to go through the changes and also tried to understand the logic for state and coloring. 

My takeaway is that we really need to rethink our handling of the coloring and the categorization of column names. As in: we need to split the two problems and handle them separately. At the moment, category assignment and also the "is categorized" status of columns is tracked via their style color attributes in the DOM, and this is also reflected in the function and variable naming throughout the app ("paintTable", "paintingData"). From what I can see, routing everything through the color requires a very considerable amount of code complexity (regex parsing of our CSS stylesheet, specific lifecycle hooks, comparing colors to infer categorization status), some of it undesirable (direct editing of DOM elements to set color rather than letting Vue handle it). 

Our main goal is to get users to tell us what categories each column and value belong to. The colors are just a visual guide / helper and should not force us to adopt complex design. I think if we dropped colors alltogether, we could probably build a simple mapping of column names to categories just via the row-click events of our tables in a few lines of intuitive code. 

From there we can think about applying color. If I understand things right, coloring the row tables is a bit tricky because we are using bootstrap and they are opinionated about colors and themes. I saw that in the current master (before this PR), you had used `:tbody-tr-class` to programmatically style the table rows. If that works, I think we should figure out the categorization, have a mapping of categories to css classes (hardcoded), and then feed the appropriate array to `:tbody-tr-class` every time a categorization update has happened. If `:tbody-tr-class` is not a viable solution, then we should look into `b-table-simple` instead of `b-table`. It is a bit more complex and less adaptable, but we gain fine-grained control over all elements of the table and if that allows us to lose complexity elsewhere, I think it's still worth it.

So my suggestion is: 
- let's split categorization from coloring, 
-  ignore the coloring for now, 
- figure out fundamentally the categorization and state keeping for the categorization, 
- refactor variables and function names to unambiguously identify them as about categorization 
- remove code about coloring aggressively and focus on reducing complexity
- once the categorization is working (with or without coloring), think about how we can style the tables with color

This may make sense to do in a separate PR, in which case I'd say we close this one and have a chat. Let me know what you think.

Issue #74, opened 2022-03-18
Decide on minimal code style


surchs:
Now that we're working on overlapping pieces of code, we need to adapt a minimal code style so our commits don't become spammed by fluffy formatting diffs. 

I suggest we agree on 
- 4 space indentation (not tab)
- emtpy newline at the end of the file

anything else? Mostly focused on avoiding useless diffs for now.

jarmoza:
One thing that could fall into this category. We should have semicolons at the end of every JS statement and not rely on automatic semicolon insertion.

surchs:
OK, thanks for pointing this out! I think we should just get this over with while we're add it and just adapt an actual (existing) code style and set up a linter and formatter for it.

PR #75, opened 2022-03-21
[STYLE] add ESLint and prettier

surchs:
This adds:
- ESLint as a code linter
- vue specific ESLint rules and the corresponding plugins
- config for ESLint and ignore file
- prettier as a code formatter
- dependencies for all of this, plus dependencies for editorconfig (needs npm install in vscode to work)

This is related to what we discussed in #74.
I think there is a tradeoff between having something going now and immediately adopting a very strict style guide. I prepared some rules that I think would make sense to adopt eventually (e.g. airbnb style guide) but commented out for now. There are no linting checks to run the code, so all linting warning and errors can be ignored for now.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,7 @@
+/dist
+/src-capacitor
+/src-cordova
+/.quasar
+/node_modules
+.eslintrc.js
+babel.config.js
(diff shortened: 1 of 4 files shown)


surchs:
@jarmoza as we had discussed on slack and #76, I'll merge this so we can use the linter in the open PRs

PR #76, opened 2022-03-21
[FEAT] add explanation component

surchs:
- [x] adds components for each of the currently used categories and links them from the annotation page.
- [x] adds an explanation sub-component to display text explanations during annotation

### DIFF
--- a/components/annot-age.vue
+++ b/components/annot-age.vue
@@ -1,12 +1,17 @@
 <template>
   <div>
-    Age
+    <annot-explanation :explanation="explanation"></annot-explanation>
   </div>
 </template>
 
 <script>
 export default {
   name: "annotate_age",
+  data() {
+    return {
+      explanation: 'This is an explanation for how to annotate age',
+    }
+  },
 }
 </script>
 
--- a/components/annot-age.vue
+++ b/components/annot-age.vue
@@ -1,0 +1,15 @@
+<template>
+  <div>
+    Age
+  </div>
+</template>
+
+<script>
+export default {
+  name: "annotate_age",
+}
+</script>
+
+<style scoped>
+
+</style>
--- a/components/annot-age.vue
+++ b/components/annot-age.vue
@@ -6,15 +6,13 @@
 
 <script>
 export default {
-  name: "annotate_age",
+  name: 'AnnotAge',
   data() {
     return {
       explanation: 'This is an explanation for how to annotate age',
-    }
+    };
   },
-}
+};
 </script>
 
-<style scoped>
-
-</style>
+<style scoped></style>
--- a/components/annot-age.vue
+++ b/components/annot-age.vue
@@ -6,10 +6,10 @@
 
 <script>
 export default {
-  name: 'AnnotAge',
+  name: "AnnotAge",
   data() {
     return {
-      explanation: 'This is an explanation for how to annotate age',
+      explanation: "This is an explanation for how to annotate age",
     };
   },
 };
--- a/components/annot-explanation.vue
+++ b/components/annot-explanation.vue
@@ -15,21 +15,19 @@
 
 <script>
 export default {
-  name: "sub-explanation",
+  name: 'AnnotExplanation',
   data() {
     return {
-      card_title: "Explanation",
-    }
+      card_title: 'Explanation',
+    };
   },
   props: {
     explanation: {
       type: String,
-      default: "No explanation has been provided yet."
+      default: 'No explanation has been provided yet.',
     },
-  }
-}
+  },
+};
 </script>
 
-<style scoped>
-
-</style>
+<style scoped></style>
--- a/components/annot-explanation.vue
+++ b/components/annot-explanation.vue
@@ -1,0 +1,35 @@
+<template>
+  <div>
+    <b-card no-body class="mb-1">
+      <b-card-header header-tag="header" class="p-1" role="tab">
+        <b-button block v-b-toggle.accordion-explanation variant="info">{{ card_title }}</b-button>
+      </b-card-header>
+      <b-collapse id="accordion-explanation" accordion="my-accordion" role="tabpanel">
+        <b-card-body>
+          <b-card-text>{{ explanation }}</b-card-text>
+        </b-card-body>
+      </b-collapse>
+    </b-card>
+  </div>
+</template>
+
+<script>
+export default {
+  name: "sub-explanation",
+  data() {
+    return {
+      card_title: "Explanation",
+    }
+  },
+  props: {
+    explanation: {
+      type: String,
+      default: "No explanation has been provided yet."
+    },
+  }
+}
+</script>
+
+<style scoped>
+
+</style>
--- a/components/annot-explanation.vue
+++ b/components/annot-explanation.vue
@@ -15,16 +15,16 @@
 
 <script>
 export default {
-  name: 'AnnotExplanation',
+  name: "AnnotExplanation",
   data() {
     return {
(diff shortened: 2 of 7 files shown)


components/annot-age.vue
```diff
@@ -0,0 +1,20 @@
...
+    <annot-explanation :explanation="explanation"></annot-explanation>
+  </div>
+</template>
+
+<script>
+export default {
+  name: "annotate_age",
+  data() {
+    return {
+      explanation: 'This is an explanation for how to annotate age',
```
jarmoza, line 12: Let's consider consistent use of single vs double quotes, especially now re: a JS linter. Thus far I have been using double quotes for strings and alternating to single quotes when quoting something within double-quoted strings, e.g. tag attribute values.

components/annot-gender.vue
```diff
@@ -0,0 +1,20 @@
+<template>
+<div>
+  <annot-explanation :explanation="explanation"></annot-explanation>
+</div>
+</template>
+
+<script>
+export default {
+  name: "annotate_gender",
```
jarmoza, line 9: Let's now be consistent with the term 'sex' or 'gender'. What is the consistent usage 
in the neuroimaging community? What is typically coded? Is 'gender' a term that's actually 
being coded for or is it actually biological sex?

components/annot-explanation.vue
```diff
@@ -0,0 +1,35 @@
...
+          <b-card-text>{{ explanation }}</b-card-text>
+        </b-card-body>
+      </b-collapse>
+    </b-card>
+  </div>
+</template>
+
+<script>
+export default {
+  name: "sub-explanation",
```
jarmoza, line 18: Should we be providing name fields for all components or just pages? This question, of course, addresses all the components in this commit. But regarding this one -

Why is the `annot-explanation` component's name just 'sub-explanation'? If we're giving names to components,
how can we make consistent naming, and with regard to `annot-explanation` bring that naming in line with
the names for the other annotation page components?

pages/annotation.vue
```diff
@@ -1,68 +1,58 @@
...
+
+    ...mapState([
+      "pageData",
+      "pageOrder"
+    ])
+  },
+}
 </script>
 
 <style scoped>
```
jarmoza, line 56: We should discuss how to bring the annotation page in line with the rest of the tool now. Do we want to address tool style consistency here? Or in a separate PR?

pages/annotation.vue
```diff
@@ -1,68 +1,58 @@
...
+// Fields listed in mapState below can be found in the store (index.js)
+import { mapState } from "vuex";
+
+export default {
+  name: "Annotation",
+  data() {
+    return {
+      // TODO: "pages" is used as static testing data. Should be replaced with actual list of used categories
+      // from global state / store
+      pages: [
```
jarmoza, line 38: We can commit this for now but should we be adding this to the store this week after these PRs?

jarmoza (changes requested):
Here we go with the annotation page! This PR is nice and simple, but I've included a few comments addressing some light changes - and some 'might as well do this now'  changes - now that we're integrating your work into the codebase.

components/annot-age.vue
surchs: Hey, thanks for the comment. That's a good point. How do you feel about just merging the https://github.com/metaneuro/annotation_tool/pull/75 PR first and then applying all that? I can do it manually here but I think that'd be faster.

components/annot-gender.vue
surchs: I think that's a good question. I think "sex" is the general standard and also [used in BIDS](https://bids-specification.readthedocs.io/en/stable/99-appendices/14-glossary.html#sex-columns), but I think that in the vast majority of cases, what we encode with this is the self-reported gender of participants. That said, I think this is something we shouldn't just decide on our own. And so for now, it might make sense to go with "sex" everywhere and keep this in mind for a future revision.

components/annot-explanation.vue
surchs: Good point. This was an oversight. I am not sure what uses the `name` attribute here. But it would make sense to me to have `name` = `fileName`. Do you agree?

pages/annotation.vue
surchs: What style do you mean? Coding style as in https://github.com/metaneuro/annotation_tool/pull/75 or css style?

pages/annotation.vue
surchs: Up to you. I thought getting them in as is for now might be easier. But if you want we can also hook them up directly in this PR. The benefit of handling this in a separate PR might be that we won't have as many merge conflicts for the downstream PRs.

components/annot-age.vue
jarmoza: That's fine.

jarmoza: That's fine.

components/annot-gender.vue
jarmoza: Sounds good.

components/annot-explanation.vue
jarmoza: Let's just do some quick research on this `name` property. It was recommended in the Nuxt tutorials for pages -from what I recall.

pages/annotation.vue
jarmoza: CSS style. Sorry - style tag seemed the best place to put this comment. Though in retrospect this kind of thing can go in the general 'Request Changes' comment box.

pages/annotation.vue
jarmoza: Separate PR it is.

pages/annotation.vue
jarmoza: I think separate issue and then PR is best for the CSS changeover. Like the gluing of the store to the data used by these components.

components/annot-explanation.vue
jarmoza: https://v2.vuejs.org/v2/api/?redirect=true#name

> #name
> 
> Allow the component to recursively invoke itself in its template. Note that when a component is registered globally with Vue.component(), the global ID is automatically set as its name.
> 
> Another benefit of specifying a name option is debugging. Named components result in more helpful warning messages. Also, when inspecting an app in the [vue-devtools](https://github.com/vuejs/vue-devtools), unnamed components will show up as <AnonymousComponent>, which isn’t very informative. By providing the name option, you will get a much more informative component tree.

jarmoza: And it is an option for the linter: https://eslint.vuejs.org/rules/require-name-property.html

jarmoza: `name` = `filename` works. I don't see a compelling reason for something more descriptive than that.

jarmoza: `name` = `filename` works. I don't see a compelling reason for something more descriptive than that.

surchs: OK, will add these then!

components/annot-age.vue
surchs: OK, PR #75 is merged. I will adopt the airbnb style with single quote strings here now.

surchs: Oh, yeah. Happy to go with that either, main goal was consistency with these commits. I went with single quotes because that's what the airbnb style guide happens to do (in case we adopt this) and because they look nicer to me. Let me know if you'd prefer double quotes for all JS strings and I'll change it.

jarmoza:
With the reversion back to double quotes, this PR is ready to be merged.

PR #77, opened 2022-03-21
[FEAT] display categorized columns on the annotation page

### LINKED ISSUE #57: Create component to show all columns mapped to a category for annotation page
When I am on the annotation page and viewing e.g. the "Diagnosis" annotation view, I want to have a summary of all the columns that I had tagged to be "about" diagnosis in the categorization step, so that I can see if any of them don't make sense and go back to the categorization step to correct this.

**Input**:
- a pre-filtered list of all column names (and if available, their descriptions from the data dictionary) that have previously assigned to the current category

**Behaviour**:
- [ ] the component shows me a compact view of the columns I have assigned to the currently shown category
- [x] the component doesn't cover more than 1/3rd of vertical viewport space or else has the option to collapse
- [x] the component has some functionality to remove a column that was added in error. This could be either something that happens on the component and changes global annotation state. Or it could be a button that brings the user back to the categorization page

TODO
- [ ] document the format that the pre-filtered list of columns is supposed to be passed to the component
- [ ] document how the component retrieves or gets passed data from the (global) state on category tags and so on

surchs:
In this PR:
-  added component to display categorized columns
-  renamed category sub-pages on annotation page with "category" prefix
-  linked new component form category sub-pages and passed the required props
-  hard-coded an example list of categorized components in the annotation page and passed it to the category sub-pages. Should be replaced by global store
-  event emitters are included for when a previously categorized column is unassigned from this category via the new component interface

Resolves #57


components/annot-age-values.vue
```diff
@@ -0,0 +1,204 @@
+<template>
+  <div>
+    <b-card no-body>
+      <b-card-header>I am a header</b-card-header>
+      <b-card-body style="position:relative; height:300px; overflow-y:scroll;">
```
jarmoza, line 5: Move to style tag below for this class or create an id for this element to reference in CSS

components/annot-age-values.vue
```diff
@@ -0,0 +1,204 @@
...
+      // The order matters here because the capture groups will match from first to last and stop on a match
+      return new RegExp(this.regularExpressions.join("|"))
+    },
+    relevantColumns() {
+      //  Return only those columns that are annotated with the current category
+      return Object.entries(this.columns)
+        .filter(([columnName, categoryName]) => categoryName === this.activeCategory)
+        .map(element => element[0]) // return only the column name that was assigned to this.activeCategory
+    },
+    filteredTable() {
```
jarmoza, line 44: See my general comment on complex return values.

components/annot-age-values.vue
```diff
@@ -0,0 +1,204 @@
...
+          return Object.fromEntries(
+            Object.entries(row)
+              .filter(
+                ([columnName, rowValue]) =>
+                  this.relevantColumns.includes(columnName))
+          )
+        }
+      )
+    },
+    uniqueValues() {
```
jarmoza, line 56: See my general comment on complex return values

components/annot-age-values.vue
```diff
@@ -0,0 +1,204 @@
...
+        this.relevantColumns.map(colName => {
+            const uniques = new Set(
+              this.filteredTable.map(row => row[colName])
+            )
+            return [colName, Array.from(uniques)]
+          }
+        )
+      )
+    },
+    unique_table_data() {
```
jarmoza, line 68: See my general comment on complex return values

components/annot-age-values.vue
```diff
@@ -0,0 +1,204 @@
...
+
+        case "isoyear":
+          // TODO: think of a way to get the values parsed without having to call the detectAge method again here
+          const ageFormats = this.detectAgeFormat(value)
+          const yearValue = parseInt(ageFormats.isoyear.replace("Y", ""))
+          const monthValue = Object.keys(ageFormats).includes("isomonth") ? parseInt(ageFormats.isomonth.replace("M", "")) / 12 : 0
+          return `${yearValue + monthValue}`
+      }
+    },
+    applyTransform() {
```
jarmoza, line 161: See my general comment on complex return values

components/annot-age-values.vue
jarmoza, line 68: See my general comment on complex return values

components/annot-age-values.vue
jarmoza, line 56: See my general comment on complex return values

components/annot-age-values.vue
jarmoza, line 44: See my general comment on complex return values

components/annot-age-values.vue
jarmoza, line 5: Move to style tag below for this class or create an id for this element to reference in CSS

components/annot-age-values.vue
```diff
@@ -0,0 +1,204 @@
...
+
+<script>
+
+export default {
+  name: "sub-numeric-validation",
+  data() {
+    return {
+      // TODO turn the range into an argument for the component
+      unique_range: {start: 0, end: 3},
+      regularExpressions: [ // using named capture groups
```
jarmoza, line 21: It's best to heavily document regular expressions when they are created. This way when you or someone new revisits the regex code, all pieces of the expression are immediately available without visual parsing of them.

components/annot-columns.vue
```diff
@@ -0,0 +1,59 @@
+<template>
+  <b-card no-body>
+    <!--    TODO: make this also toggleable like the explanation tab-->
+    <b-card-header>Review the annotated columns</b-card-header>
+    <b-card-body style="height:30vh; overflow-y:scroll;">
```
jarmoza, line 5: Move to style tag below for this class or create an id for this element to reference in CSS

components/annot-columns.vue
```diff
@@ -0,0 +1,59 @@
...
+      <b-list-group>
+        <b-list-group-item
+          class="d-flex justify-content-between align-items-center"
+          v-for="[columName, categoryName] of Object.entries(filteredColumns)">
```
jarmoza, line 9: Why is Object.entries performed twice here?

components/annot-columns.vue
```diff
@@ -0,0 +1,59 @@
...
+        }
+      )
+    }
+  },
+  computed: {
+    filteredColumns() {
+      // "columns" is an object of the form
+      // { columnName: "CategoryName", ... } that contains all annotated columns and the categories they were mapped to
+      // In normal use, this object is kept in the global state and is passed here via the parent components
+      return Object.fromEntries(
```
jarmoza, line 43: This is compact, but can you please write this in a few steps? And with comments on the process?

(Maybe we should put a limit on the levels deep of complexity a return value can have before we write the steps of the code separately.)

components/category-age.vue
```diff
@@ -0,0 +1,43 @@
...
+  </div>
+</template>
+
+<script>
+export default {
+  name: "annotate_age",
+  data() {
+    return {
+      explanation: 'This is an explanation for how to annotate age',
+      activeCategoryName: 'Age'
```
jarmoza, line 26: I presume this is a store value? (Likewise across the other `category-*` components.)

jarmoza (changes requested):
Okay. So overall, functionally, the pieces in this PR look good. I didn't notice any errors. Obviously, there are pieces that will be fleshed out more in the future, bits that need to be hooked to the store, and aesthetic changes as discussed. So most of my comments below and specific to the files have to do with config and coding practices.

And nicely done for your first complex piece of Vue code!

General comments:

1) Now that components are getting more complicated, we should use nested directories within the components directory. Though initially this would require the component name to include a concatenated version of the subdirectory path in its name, this can be overcome by placing the path in `components.dirs` in `nuxt.config.js`. See: https://nuxtjs.org/docs/directory-structure/components/#nested-directories

2) One of the reasons I heavily comment code is through lessons learned from past development and collaborative projects where either I forgot what code was doing or people needed to quickly know without parsing my code what it is intended to do. Overall, even though it does not change the functionality I would suggest taking some time now while the code is relatively fresh in your mind to go through all of these new files in this commit and provide some comments - particularly for the functions that are producing values used by and shared between components in the interface.

3) Complex return values (see examples `filteredTable`, `uniqueValues`, `unique_table_data`, `applyTransform` in `annot-age-values.vue` and `filteredColumns` in `annot-columns.vue`):
      - Though it's tempting to use this kind of syntax to produce a return value in one line, code like this becomes confusing for people who are approaching it. I would recommend breaking down this return value creation process into several lines before the final value is returned.
      - As I mention in one comment in `annot-columns.vue`, maybe we should institute some rule for the number of levels deep a (return) statement can be before it should be broken up into several distinct steps in our code.
 
3) Given our discussion re: semicolons, this would be a good chance to insert them in this PR.

4) Consider blank lines between distinct pieces of functionality. They can make code more readable, if a little less compact.

components/annot-columns.vue
jarmoza, line 9: I'm getting the following warning from here:

>  WARN  in ./components/annot-columns.vue?vue&type=template&id=6b92ec6b&        friendly-errors 10:51:55
> 
> Module Warning (from ./node_modules/vue-loader/lib/loaders/templateLoader.js): friendly-errors 10:51:55
> (Emitted value instead of an instance of Error) <b-list-group-item v-for="[columName, categoryName] in Object.entries(filteredColumns)">: component lists rendered with v-for should have explicit keys. See https://vuejs.org/guide/list.html#key for more info.

jarmoza (changes requested):
Just noticed a warning coming from `annot-columns.vue`, around line 10.

surchs:
Thanks @jarmoza for the detailed review. That's very helpful!

> we should use nested directories within the components directory

I had used that first for the same reasons you mention. The reason I changed back to a flat list was that I read a [convincing piece on best practice](https://vueschool.io/articles/vuejs-tutorials/how-to-structure-a-large-scale-vue-js-application/) arguing against nested components. The bottom line was: 
- flat dir gives you quick (visual and filter) access to components, no need to navigate dirs up and down
- what namespace control you get from nested dirs, you can also get from kebab-case naming the component itself. Vue treats both the same
- the components are (by default) alphabetically sorted, so stick together contextually

There are some other points in the piece, give it a read. What convinced me most was that it gets pretty annoying to navigate all these sub-dirs (how deep should you go) and then you don't know where a new component should go. So I'd vote for keeping things flat until we find that this suggestion doesn't work for us.

agreed on the other points regarding commenting and more readable constructs for the returned values. As we discussed, I wonder if this means we should not use arrow functions for things that are more than one or two levels deep.

jarmoza:
@surchs Just a note that since the #76 merge there are now merge conflicts for this branch that must be reconciled.

surchs:
OK, @jarmoza, as we had just discussed, I turned your unaddressed comments into new issues so we can address them after the demo but merge the functional code now to get ready for the demo. I'll merge this now as is.

PR #86, opened 2022-03-24
Creating a layout for the tool and moving navbar to the layout

jarmoza:
A layout vue template has been created in `layouts/default.vue` and `tool-navbar` has been moved outside of the pages and into that default layout.

As such pages are no longer responsible for generating the component and only have to worry about setting the current page via store function `setCurrentPage`.

The store now maintains a member `currentPage` for the navbar component.

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -1,0 +1,38 @@
+<template>
+  
+    <div>
+
+        <!-- Navigation bar -->
+        <tool-navbar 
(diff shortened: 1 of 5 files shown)


surchs (approved):
Awesome @jarmoza, thanks for the PR. I think this makes things look very nice and clean! Confirmed that it works and agreed with the changes to the code

PR #87, opened 2022-03-24
[FEAT] add discrete value annotation

surchs:
This PR adds a drop-down based functionality for mapping discrete values to a set of available choices.
For now we use this only for the sex category

### DIFF
--- a/components/annot-discrete-choices.vue
+++ b/components/annot-discrete-choices.vue
@@ -6,6 +6,8 @@
         <b-table :items="displayTable" :fields="exampleFields">
           <template #cell(select_an_appropriate_mapping)="row">
             <!--Bootstrap-Vue doesn't have a great option here so I am using https://vue-select.org/-->
+            <!--Note: we use the $event statement to add the row data to the payload of the @input
+                event without replacing the original event payload -->
             <v-select
               label="Standard"
               :options="annotationOptions"
@@ -13,6 +15,12 @@
             ></v-select>
           </template>
         </b-table>
+        <b-button
+          variant="success"
+          @click="applyTransform"
+          :disabled="buttonDisabled"
+          >Confirm and Upload
+        </b-button>
       </b-card-body>
     </b-card>
   </div>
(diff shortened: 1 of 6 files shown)


surchs:
ok @jarmoza, I think this works now as we would like it to. Ready for your review!

jarmoza:
@surchs Just tested this. I added categorized as 'sex' the actual 'sex' column and two other columns. Proceeding to annotation, I mark the 'M' as 'male' and the 'F' as 'female' and the blank value as 'missing value'. `Confirm and Upload` remains disabled as expected – because the other columns have not been annotated with accepted 'sex' values.

However, when I removed the miscategorized columns the button remains disabled.

One good point on that last action is that those miscategorized columns are actually unlinked from the 'sex' category on the categorization page when I navigate back to it.

jarmoza:
@surchs Let me know if you have any thoughts on my recent comment. The easiest way would be to utilize the `download` page `accessible` flag in `pageData` similar to the previous pages. 

What could/should be happening is that after a function that could change the page state - i.e. clear to proceed to the download page - a call should be made to change the page's state and in turn the actions that should follow that particular state change.

surchs:
@jarmoza: what data did you test this with? One of our example files?

surchs:
@jarmoza: I can confirm that this does not currently work:

1. Label two columns as "about" sex
2. Go to the annotation page and select the "sex" tab
3. using the annotated-columns component, delete one of the two columns
4. label all the unique elements of the remaining column
5. expect the button to become enabled -> it does not

The problem was that the method to enable / disable the button did not react to the number of columns changing and was still checking whether the unique values of the deleted column would be annotated (which they of course could never be), thus being forever stuck in a "not all annotated" state.

One fix for this is to just recreate the valueMapper that is being checked to determine the annotation status with a watcher whenever the columns change. This works, in the sense that now we determine the annotation status only based on the currently selected columns.

However, we now have another problem: because we are using the v-select dropdown component inside scoped slots, and because v-model does not work inside scoped slots (see [this stackoverflow](https://stackoverflow.com/questions/55211175/using-v-model-inside-scoped-slots/55212045#55212045) and the linked issue from vue) we are using the `@input` event directly to react to a dropdown value being selected. The selected value is still modeled with `v-model` inside the dropdown component, but no longer tied to the state of the app. So when we re-initialize the valueMapper as empty via the watch function by removing a column, an already selected value in the dropdown does not get reset as well. So you could end up with the very confusing situation where all dropdown values have some string selected but the button is disabled because internally the mapping is emtpy.

This does not happen if you navigate away from the page and come back, but that's no good. So in short, I think we need to not throw out the valueMapper and re-initialize as emtpy, but rather pick and choose what should remain and what should be removed.

surchs:
OK, thanks @jarmoza. I will merge this now then.

PR #91, opened 2022-03-28
Adding download page and hooking it into annotation tool workflow

jarmoza:
This PR adds a basic download page that follows the annotation page. 

Currently it is always accessible from the annotation page (not unlocked upon completion of annotation - like the other pages). This is just for demo purposes. Going to the download page while in the annotation page without having 'annotated' any data merely produces a blank page with "download" button.

However, once an annotation has occurred, going to the download page produces a table of the "annotated" data.

NOTE: Some other aspects of this implementation should be revisited after the demo tomorrow - including how/when annotation status is saved in the store. This will include moving replicated functionality like the 'Confirm and Upload' button upwards out of the annotation page's components.

### DIFF
--- a/components/category-age.vue
+++ b/components/category-age.vue
@@ -12,7 +12,6 @@
       :active-category="activeCategoryName"
       :columns="columns"
       :data-table="dataTable"
-      :pageData="pageData"
       @update:dataTable="$emit('update:dataTable', $event)"
     ></annot-age-values>
   </div>
@@ -34,11 +33,7 @@ export default {
     },
     dataTable: {
       type: Array
-    },
-    pageData: {
-      type: Object,
-      required: true
-    }    
+    }   
   }
 }
 </script>
--- a/components/category-age.vue
+++ b/components/category-age.vue
@@ -12,6 +12,7 @@
       :active-category="activeCategoryName"
       :columns="columns"
       :data-table="dataTable"
+      :pageData="pageData"
       @update:dataTable="$emit('update:dataTable', $event)"
     ></annot-age-values>
   </div>
@@ -33,7 +34,11 @@ export default {
     },
     dataTable: {
       type: Array
-    }
+    },
+    pageData: {
+      type: Object,
+      required: true
+    }    
   }
 }
 </script>
--- a/pages/annotation.vue
+++ b/pages/annotation.vue
@@ -3,41 +3,40 @@
     <b-container fluid>
 
         
-            <!--
-                TODO: revisit the client-side render solution or remove this comment
-                The v-for statement below was causing a mismatch between client-side and server-side
-                DOM. In particular, the first element in "pages" (Age) was rendered twice. The error message was:
-                Vue warn]: The client-side rendered virtual DOM tree is not matching server-rendered content.
-                This is likely caused by incorrect HTML markup,
-                for example nesting block-level elements inside <p>, or missing <tbody>.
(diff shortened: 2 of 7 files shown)


surchs:
@jarmoza: Thanks for this PR! I only have one comment: why do we need to force-navigate to the download page every time we update the dataTable? With the demo in mind, that'd make you jump back and forth a lot (e.g. look at age -> "upload/confirm" -> be brought to download page -> navigate back to annotation -> ...).

I think it should be enough to just enable the download page once we have made one dataTable update emit and then let the user navigate to the download page on their own.

jarmoza:
> why do we need to force-navigate to the download page every time we update the dataTable

Well, we don't _need_ to do that. It's a UI choice. Ideally, we'd want some sort of feedback from the upload button. The most immediate choice was to have the user move along in the annotation process (e.g. to the download page). Currently, there is no feedback as to what's happened otherwise.

Another possibility is to mimic the UI of the previous pages and to enable a 'Next' button once the user clicks the 'Confirm and Upload' button.

surchs:
Yeah, I think I'd prefer if we just make the download page accessible conditional on dataTable.annotated not being null / the annotation page having received at least one dataTable update event

jarmoza:
Okay. I'll add the button then.

jarmoza:
@surchs Okay. Take a look and see what you think.

jarmoza:
Additional note: Oh - I had tried to put the top part of the annotation page in a row - thus the indented top part, but row styling made it shrink in width. So the download page button is just in its own row beneath the annotation UI.

pages/annotation.vue
```diff
@@ -1,39 +1,63 @@
...
+
+            <!-- Button to proceed to the next page -->
+            <!-- Only enabled when at least one annotation table write has been done -->
+            <b-col cols="3">
+                <b-button
+                    class="float-right"
+                    :disabled="!pageData.download.accessible"
+                    :to="'/' + pageData.download.location"
+                    :variant="nextPageButtonColor">
+                    Next step: Categorize columns
```
surchs, line 52:
```suggestion
                    Next step: Review and download the harmonized data
```

components/category-age.vue
```diff
@@ -12,6 +12,7 @@
       :active-category="activeCategoryName"
       :columns="columns"
       :data-table="dataTable"
+      :pageData="pageData"
```
surchs, line 15: see other pageData comment

surchs (commented):
@jarmoza: works well, I like it.

One comment, one suggestion
- suggestion: let's call the button "Download" (I suggested inline)
- comment: let's remve the pageData props (see comment)

With these two, looks good to go!

surchs:
> thus the indented top part

OK, let's squash the PR then so we don't get confused by all the styling commits

pages/annotation.vue
jarmoza: > suggestion: let's call the button "Download" (I suggested inline)

Oops! Left over from my copy-paste from the landing page. Will replace.

> comment: let's remve the pageData props (see comment)

Good catch! Use of `pageData` here is leftover from a previous implementation where it was necessary as a prop for components since the 'Confirm and Upload' button(s) reside inside them – and was being used to proceed to the download page. Passing it through the annotation components is no longer necessary.

surchs (approved):
great. looks good to me! :tada:

Issue #94, opened 2022-03-28
dataTable updating must not conflict between categories


surchs:
currently we are only passing dataTable.original but components must be aware of previous annotations, so dataTable.annotated must also be passed.

Issue #96, opened 2022-03-28
dataTable.annotated should initialize as a copy of dataTable.original


surchs:
Because of #94 we need to make sure that dataTable.annotated is not null when we first load the annotation page so we can pass it down as a prop to our annotation components.

This also affects the current pageAccess state for the download page so we will have to find an alternative solution to detect when the download page should become accessible.

jarmoza:
Fixed as of #97 . Closing this issue.

PR #97, opened 2022-03-28
[FIX] pass the original and annotated dataTable to components

surchs:
- Components get as props both the original dataTable and the annotated dataTable
- inside components the updated mappings of values are applied to the annotated dataTable in order not to lose previous annotations
- raw values are still displayed based on the original dataTable.

Fixes: #94

### DIFF
--- a/components/annot-age-values.vue
+++ b/components/annot-age-values.vue
@@ -191,7 +191,7 @@ export default {
   },
   props: {
     dataTable: {
-      type: Array,
+      type: Object,
       required: true
     },
     activeCategory: {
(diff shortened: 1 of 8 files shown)


surchs:
@jarmoza: this won't work as is until #96 is addressed. Until then, you can use a quickfix by replacing `mounted` on the `annotation.vue` page with:

```javascript
    mounted() {

        // 1. Set the current page name
        this.$store.dispatch("setCurrentPage", "annotation");

        // TODO: remove this
        if (this.dataTable.annotated === null) {
          console.log(" I am setting the annotated DataTable to be a copy of the original now!")


          this.$store.dispatch("saveAnnotatedDataTable", this.dataTable.original);
        }
    }

};
```

jarmoza:
I can confirm that this is working. My recent commit also clones the original data table when it is uploaded and saves the clone as the annotated table (see `setDataTable()` mutation in the store).

jarmoza (approved):
I can confirm that this is working.

Issue #93, opened 2022-03-28
Styling improvements


surchs:
We want some styling improvements

- [x] on the annotation pages, the upload-and-confirm buttons should be called "Save Annotation"
- [x] these buttons should also be grey when deactivated, consistent with the "Next Step" button style
- [x] all template cardnames (I am a header ...) should be replaced with something meaningful
- [x] cards on the annotation page should have some vertical spacing
- [x] all tables should use the striped-style
- [x] the tool name should be changed from "Origami ..." to "Metabagel (?) Annotation Tool"
- [x] color on the annotation tabs should match the category colors (optional)
- [x] check if we can make better use of whitespace in the annotation app to make things look "nice" (optional)

jarmoza:
One note here while I work on this.

> all tables should use the striped-style

Striped tables break the table painting on the categorization page. So this is a no-go there.

surchs:
OK, could we just use them on the annotation page? If not, go with whatever is easiest here!

jarmoza:
I believe all of these style improvements have now been addressed as of #108. @surchs Can you confirm?

jarmoza:
Closing this issue as everything in it that will be addressed has been addressed.

PR #100, opened 2022-03-28
Styling improvements for demo

jarmoza:
This implements all of the changes as suggested in #93 - though there are still some places for whitespace improvements on the annotation page.

Also included is the stretching of the download page table to full size of the browser window.


surchs (approved):
This is looking very good @jarmoza. Great stuff, let's merge it!

PR #108, opened 2022-04-14
Post demo cleanup - Annotation page component overhaul and formatting

jarmoza:
**General** 

1. All UI text has been placed in a `uiText` object in the data of their respective component
2. Fields from the store are given to annotation components and subcomponents via provide/inject mechanism. Other fields are passed as props.
3. All props now use the object notation with `type` and `required`. If a prop is given to a component but is not used, it is listed but not marked as `required: true`.
4. All entries in the script section of Vue template files have been arranged according to the Vue style guide: props, provide/inject, data, computed, Vue lifecycle hooks, methods. Within each group all fields, methods, etc. are alphabetized.
5. Functional programming solutions (e.g. Object.fromEntries(), entries(), map(), etc.) remain where useful. However, instances of multiple-levels-deep functional programming were reevaluated with two concerns in mind. 1) Readability and 2) efficiency - with the latter taking precedence over the former. Cases where both readability and computational efficiency could be improved were changed to an imperative approach.
6. Formatting consistency with four space indents and line spacing has been added. A few other bits of tidying - like capitalizing the beginning of comments and the addition of a `p_` hungarian notation to distinguish parameters from local variables. Some of this is just for consistency's sake for the time being until we introduce a prettifier - which will be coming up very soon.
7. All instances of '==' and '!=' have been replaced with '===' and '!==' respectively.

**`index.js`**

1. Corresponding action/mutation functions for the new `annotationDetails` field have been added to the store. This field contains information for generating the individual annotation tabs. This field holds the same purpose as the former `pages` field in `annotation.vue`, but is now moved to the more central place of the store where it can be properly initialized via config (in the future). Each page object also contains a new `options` field for any specialized parameters needed for any of the tabs. The field is initialized in the `nuxtServerInit` action just like the categories.
2. Corresponding action/mutation functions to revert annotation changes if a column unlinked from its category post-annotation has been added to the store,`revertColumnToOriginal`. This is currently called when a `remove` button is clicked in the `annot-columns` component.
3. New getters have been added to the store to retrieve column and value descriptions from the data dictionary: `columnDescription` and `valueDescription`.
4. Function `convertTsvLinesToDict` in the store has been renamed to `convertTsvLinesToTableData` in order to better reflect its purpose. But more importantly, a bug in this function was fixed that was still allowing
blank lines in files to enter the data table.

**`category-*`**

1. `category-*.vue` files have all been removed. They have been replaced with the more general `annot-tab` component.

**`annot-tab`**

1. `annot-tab` contains the `annot-explanation` and `annot-columns` components of the previous `category-*` components but now includes a `component` representing the data type specialized components for each category on the annotation page. The information to generate these tabs is now found in the store in the `annotationDetails` field which is initialized in the `nuxtServerInit` action.
2. Three computed properties that were previously in the specialized annotation components have been moved up into the `annot-tab` component: `filteredTable` (now `filteredDataTable`), `relevantColumns`, and `uniqueValues`

**`annot-age-values`, `annot-discrete-choices`, and `annot-vocabulary`**

1. Save annotation buttons remain in the specialized components of the annotation tabs. This is because decoupling them without messy event handling or refs to child methods was not possible. The small redundancy of each component having its own save button is preferable.
2. Function names and fields not moved to the `annot-tab` component that remain in the specialized annotation components have been standardized  (i.e. `saveButtonDisabled`, `saveButtonColor`, `applyAnnotation`, `transformedValue`, `initializeMapping`, `updateMapping`, etc.)
9. `applyTransformation` has been renamed to `applyAnnotation` and a new helper function has been introduced for the sake of uniformity: `transformedValue`. This latter function is the place where an annotation component can go to get the new value given a column and raw value. The hope was that `applyAnnotation` could be raised up to the `annot-tab` level but given some slight differences, it made sense to keep it here. In the very least, `applyAnnotation` and `transformedValue` can give these components consistent function naming. 
10. Similarly, 'mapping' function names have all been standardized - `initializeMapping`, `updateMapping`, and `uploadMapping` (future functionality withstanding).
11. As the `checkAnnotationState` functions just have the purpose of setting the disabled status of the save annotation button, that statement has been pulled out of the function. The disabled status is now more explicitly set via the new boolean return value of `checkAnnotationState` instead.

**`annot-columns`**

1. `retrieveColumnDescription` has been implemented to display the data dictionary description of the column next to its name in this component. (Can always be dropped if the screen appears to be too cluttered because of it.) This uses the new `columnDescription` retrieval method in the store
2. The `v-for` key warning has been removed by adding the key as `columnName`

**`annot-vocabulary`**

1. `getDescription` has been split into two functions and moved to the store since it contained general, reusable data dictionary functionality. These are the new `columnDescription` and `valueDescription` functions


components/annot-vocabulary.vue
```diff
@@ -1,299 +1,356 @@
...
+
+                    // 3. Trigger a save of this transformation to the annotated table in the store
+                    this.$emit("update:dataTable", {
+
+                        transformHeuristics: this.vocabularyMapping,
+                        transformedTable: transformedTable
+                    });
+                } 
+                // Else, this is 'column' mode
+                else {
```
surchs, line 258: I agree that adding the comment here is helpful as a reminder that the only other option is "column" at the moment. But in this case I think we should make the code explicit and use `else if ("column" === ...)`

components/annot-vocabulary.vue
```diff
@@ -1,299 +1,356 @@
...
+                        if ( Object.values(this.vocabularyMapping[columnName])
+                                   .some((uniqueValue) => null === uniqueValue) ) {
+
+                            columnHasUnmappedValues = true;
+                            break;
+                        }
+                    }
+                } 
+                // Else, this is in "column" mode
+                else {
```
surchs, line 286: as in the above, I would be explicit here. The fact that we add a comment is a good indicator that the code could use it

components/annot-vocabulary.vue
```diff
@@ -1,299 +1,356 @@
...
+
+                        for ( const value of this.uniqueValues[columnName] ) {
+
+                            console.log(`value: ${value}`);
+                            this.vocabularyMapping[columnName][value] = null;
+                        }
+                    }
+                } 
+                // Else, this is 'column' mode
+                else {
```
surchs, line 320: as above, would use `else if`

components/annot-vocabulary.vue
```diff
@@ -1,299 +1,356 @@
...
+                console.log("annot-vocabulary:updateMapping");
+
+                // 1. Update the mapping with the new value
+
+                // If this is 'row' mode
+                if ( "row" === this.options.mode ) {
+                    this.vocabularyMapping[p_tableRow.column_name][p_tableRow.raw_value] = p_newValue;
+                } 
+                // Else, this is 'column' mode
+                else {
```
surchs, line 346: as above, I'd suggest `else if`

components/category-select-table.vue
```diff
@@ -4,23 +4,25 @@
...
+        <b-row>
+            <p class="instructions-text">
+                {{ instructions }}
+            </p>
+        </b-row>
 
         <!-- Category selection table -->
         <b-row>
             <b-table
+                outlined
```
surchs, line 20: oh nice, they have recommended order for these things too? wow.

components/file-selector.vue
```diff
@@ -1,152 +1,159 @@
...
+                        },
+                    });
+                } 
+                // B. JSON file parsing
+                else if ( this.knownContentTypes["json"] === this.contentType ) {
 
-						// a. Save a reference to the loaded contents
-						myJson = e.target.result;
+                    // I. Reference to this json object in this component's data
+                    var myJson;
```
surchs, line 88: since you're making these things consistent, I'd use 
```suggestion
                    let myJson;
```
here

pages/categorization.vue
```diff
@@ -1,319 +1,245 @@
...
-						let columnLowercase = column.toLowerCase();
+            tableClick(p_clickData) {
 
-						// II. Try to match the json key with one in the tsv file
-						if ( headerFields.includes(columnLowercase) ) {
+                // 1. Style or unstyle table row
 
-							for ( let index = 0; index < this.columnToCategoryTable.length; index++ ) {
+                // A. Determine if category-column linking or unlinking has occurred
+                const linking = !( this.selectedCategory === this.columnToCategoryMap[p_clickData.column] );
```
surchs, line 220: Could this just be 
```suggestion
                const linking = this.selectedCategory !== this.columnToCategoryMap[p_clickData.column] ;
```
? If so, I'd find that easier to read

store/index.js
```diff
@@ -148,11 +159,71 @@ export const actions = {
...
+				id: 0,
+				category: "Age",
+				explanation: "This is an explanation for how to annotate age.",
+				options: {},
+				specializedComponent: "annot-age-values"
+			},
+			{
+				id: 1,
+				category: "Sex",
+                explanation: "This is an explanation for how to annotate sex.",
```
surchs, line 177: indent seems off

store/index.js
```diff
@@ -148,11 +159,71 @@ export const actions = {
         	"Assessment Tool"
 		];
 
+		// 0. This annotation information is default but we can swap out and reinitialize
+		// annotation data structures by calling 'initializeAnnotationDetails' with a new
+		// object containing annotation information for each category
+		const annotationDetails = [
```
surchs, line 165: cool, I like this!

store/index.js
```diff
@@ -148,11 +159,71 @@ export const actions = {
...
-		commit("setupCategories", categories);	
+		commit("setupCategories", categories);
+
+		// 2. Setup annotation-related data structures based on the given categories\
+		commit("setupAnnotationDetails", annotationDetails);
 	},
 
 	// Tool navigation
+
+	enablePage(p_context, p_navData) {
```
surchs, line 206: how about `initializePage`? Maybe not quite the right name, but this method seems to do more than just make the link clickable, particularly in the future?

store/index.js
```diff
@@ -207,15 +278,31 @@ export const actions = {
...
 
-	unlinkColumnWithCategory(p_context, p_linkingData) {
+	unlinkColumnFromCategory(p_context, p_linkingData) {
 
 		p_context.commit("removeColumnCategorization", p_linkingData.column);
 	},
 
 	// Annotation page actions
 
+	revertColumnToOriginal(p_context, p_columnName) {
```
surchs, line 290: Could you add a docstring comment to explain what this method does and when it would be called? I assume this is when someone wants to reset an annotation?

surchs (approved):
Thanks @jarmoza, this is a really excellent PR! The code got immensely more readable and clear. This is great, and great work!

I made some small comments, the main thing would probably be the eslint config to make sure we have consistent whitespace.

.eslintrc.js
```diff
@@ -1,22 +1,89 @@
...
+        browser: true,
+        es2021: true,
+    },
+
+    extends: [
+
+        // "airbnb", // TODO: activate this styleset
+        "eslint:recommended",
+        "plugin:vue/strongly-recommended",
+        // "plugin:prettier/recommended", // TODO: consider activating
```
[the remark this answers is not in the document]
jarmoza: Sounds good to me at this point. I'll remove the commented out lines.

.eslintrc.js
```diff
@@ -1,22 +1,89 @@
...
+            allowEmptyLines: true
+        }],
+
+        "vue/no-unused-vars": [ "error", {
+
+            // Ignore unused variables beginning with '_' character
+            // (Currently just used for unused variables in v-for statements)
+            "ignorePattern": "^_"
+        }]
+    }
```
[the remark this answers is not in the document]
jarmoza: @surchs I've changed these to warnings and added some specific rules about indent, vue script-indent, comma spacing, and trailing commas.

The idea here is to explicit about the rules we want to enforce even though they might be default in eslint and vue-eslint rules?

components/annot-continuous-values.vue
```diff
@@ -1,28 +1,40 @@
...
+        },
 
-<style scoped>
+        name: "ContinuousTable",
+        
+        data() {
+        
+            return {
+        
+                column: "cell(raw)"
```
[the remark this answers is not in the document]
jarmoza: Yes, that's correct. There's not much to this component currently – though it appears you still had some functionality in mind to be implemented. e.g. This is your work. The only edits I really made here was for formatting and the standards we wanted for things like `props`. As it stands, the `data` object can be removed.

components/annot-explanation.vue
```diff
@@ -1,33 +1,53 @@
...
+
+        <b-card no-body class="mb-1">
+
+            <b-card-header class="p-1" header-tag="header" role="tab">
+                <b-button block v-b-toggle.accordion-explanation variant="info">
+                    {{ uiText.cardTitle }}
+                </b-button>
+            </b-card-header>
+
+            <b-collapse id="accordion-explanation" accordion="my-accordion" role="tabpanel">
```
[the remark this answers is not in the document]
jarmoza: Should be a simple fix. I've created a new issue for this here: https://github.com/metaneuro/annotation_tool/issues/109

components/annot-tab.vue
```diff
@@ -0,0 +1,110 @@
...
+
+</template>
+
+<script>
+
+    export default {
+
+        props: {
+
+            details: { type: Object, required: true }
```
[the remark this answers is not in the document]
jarmoza: The only issue with the current implementation is that I am unsure if `is` can be validated. One solution would be to pass `specializedComponent` in a second time as String-based prop where the `validator` function can be used.

components/annot-tab.vue
```diff
@@ -0,0 +1,110 @@
...
+        },
+
+        computed: {
+
+            filteredDataTable() {
+
+                const filteredTable = this.dataTable.original.map((row) => {
+
+                    return Object.fromEntries(
+                        Object.entries(row).filter(([columnName, /*rowValue*/]) =>
```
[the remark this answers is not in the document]
jarmoza: This is actually something I just didn't revert back to uncommented.

I think this just is a matter of how we want to handle instances of this scenario. With what you wrote above, we would write a comment to include information on absent variable names.

The trick is if an unused variable occurs in the middle of an argument list, this commented out or blank arg name is actually a JS syntax error. And that's the reason I currently have the `args` option in `.eslintrc.js` set to `off`. If we retain that setting, this `rowValue` here can be uncommented out.

.eslintrc.js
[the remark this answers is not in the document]
surchs: Cool. No, my comment was about the fact that when I run eslint with this config locally, I don't get yelled at for trailing whitespace and I can chose what level of indentation I want for each line as I see fit. I think we should pick some indentation style and remove trailing whitespaces so that its consistent in the code and we don't get diffs because one of us accidentally added or removed some whitespace.

store/index.js
jarmoza: Perfect. 👌

store/index.js
surchs: Cool. But we should check why your eslint run didn't catch that.

surchs (commented):
I think you need to run one more `eslint --fix` to get the remaining errors (annotation component and some js config files). Otherwise this is good to go :tada: !

jarmoza:
@surchs Yup! I noticed. Pushed the eslint changes for the pages in a separate commit outside the PR right afterward.

PR #110, opened 2022-04-27
Initial missing values implementation

jarmoza:
I'm setting this initial version of the missing values component up as a PR. It's currently functional, but a code and functionality review is warranted.

### DIFF
--- a/components/annot-missing-values.vue
+++ b/components/annot-missing-values.vue
@@ -1,0 +1,212 @@
+<template>
+
+    <div>
+        <b-card class="annotation-card">
+
+            <b-card-header>{{ uiText.title }}</b-card-header>
+
+            <b-card-body class="missing-values-card-body">
+                <b-table
+                    striped
+                    :fields="fields"
+                    :items="tableItems">
+                    <template #cell(not_missing)="data">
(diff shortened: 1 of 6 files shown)


surchs (commented):
Thanks @jarmoza for working on this!

I think this looks good as a component. I have two comments.

1. I think the missing values component doesn't have to / should not guess what a missing value is. In most cases we will let the user tell us (only exception I can see is for numerical data). So it could be just a dumb "state viewer". Some other component tells the state that a bunch of values are "missing" and then this component know what the "relevant" columns currently are and displays for them the "missing values". Then it only has to handle the case of "nope, this is actually not a missing value after all". That way we can keep the "guess missing values" logic with the corresponding categories (e.g. age annotation) and also keep the list of "guessers" short. Another reason for this is that guessing seems pretty error prone, and I think we should avoid that complexity until we know that it is necessary.
2. relatedly, there currently doesn't seem to be a way to designate a value "missing" (i.e. you start with a list based on the guesses and can remove from there). Maybe we could just have the missing-value-component listen to the missing-value-list in the state and teach the other components to emit "this is now a missing value" events to the state.

What do you think?

PR #123, opened 2022-05-04
Implement missing values component

### LINKED ISSUE #58: Create a component to show missing data
A user annotating e.g. age on the annotation page should see a representation of all of the values we think are representing missing values in the columns that are "about" age.

**data in**:
- list of columns (and their names) previously categorized as "about" age
- the raw values from these columns
- if available, the mappings for these raw values from the data dictionary
- a list or index of unique values we think may represent missing data (this could also generated by a function inside the component)

**behaviour**:
- [ ] the component shows me every value we consider to reflect "missing data" (e.g. table, list, ...)
- [ ] together with these values, I can see the name of the column the value came from (e.g. "XX" could stand for missing data in one column but not in another)
- [ ] there is an intuitive way for me to label a value as "not missing data". e.g. with a button that is right next to the value, or a checkbox, ...
- [ ] when I complete the action to label a value as "not missing data" (e.g. press a button), it disappears from the values shown in the component

**data out**:
- a list of values and the columns they came from, that reflect "missing data"

To discuss:
- there should probably also be a way to undo this operation or declare values as "missing" in a standardized way
- where should we keep track of these missing values?

Accept when:
- [ ] in any other category annotation component (sex, diagnosis, ..., but not age) I can designate a value as missing and it will be displayed by the missing value component for the corresponding category. See #117
- [ ] in the missing value component, I can designate a value that was previously classified as missing to be not missing (e.g. remove). This will remove the value from being shown in the missing value component and will also list the value again for annotation in the corresponding annotation category component
- [ ] values displayed in this component (e.g. labeled missing) are displayed with their column name (from the table), and with any description derived from a data dictionary if available.

surchs:
This PR implements the version of the missing value component we have discussed in #58. The component does not detect or create missing values, it only does two things:

- for the currently relevant column names, show the missing values currently kept in the global store
- provide a UI element (button) to declare any of these values as "not missing". this will trigger an event chain and update the global store by removing this value from the array of missing values of its columnName

Writing to `missingColumnValues` in the store via the `setMissingColumnValues` mutation was not reactive, because Vue doesn't track indexed changes to Objects (see https://v2.vuejs.org/v2/guide/reactivity.html). I have replaced the store logic with an `Object.assign` statement. This will continue to work with the partial updates we are receiving from the annotation tab,

Example:
```javascript
let state_missingColumnValues = {
        "sex": ["na", "999"],
        "diagnosis": ["AD", "none", "dunno"] // AD is not a missing value, so we remove it
};
const new_missing_values = {"diagnosis": ["none", "dunno"]};
state_missingColumnValues = Object.assign({}, state_missingColumnValues, new_missing_values);
console.log(state_missingColumnValues);

// { sex: [ 'na', '999' ], diagnosis: [ 'none', 'dunno' ] }
```

Another thing I did here is to give the missing Value Component direct access to the store state. This is different from what we do for the other components and what was previously done, where we give only the pages (e.g. annotation) access to the store and then provide/inject or prop-chain the information down to the components. Two reasons why I did it this way here:

- the component already has state access for the getter (`valueDescription`)
- ~~none of the other components needs access to this information so provide / inject seems a bit overkill. similarly so would be props~~ edit: I just realized that this is not true. Other components will need this as well. So yeah, maybe we should use provide / inject
- it is easier to see what's going on while inside the component (this is a thing from the state).

But open to change to props or provide / inject.

Sidenotes:
- I'm not sure why I'm still getting white-space diffs here, I thought I had solved that by running eslint beforehand. Sorry about that.
- because this gets implemented before we have a way to "create" missing values, and because the missing values component no longer makes its own guesses, I stuck some temporary event emitter in the annotation page to create some missing values to debug with (targeted to our example.tsv file). I'll remove these once we have a way to add missing values!

Resolves #58

### DIFF
--- a/components/annot-missing-values.vue
+++ b/components/annot-missing-values.vue
@@ -28,7 +28,7 @@
 <script>
 
     // Allows for reference to store data by creating simple, implicit getters
-    import { mapGetters } from "vuex";
+    import { mapGetters, mapState } from "vuex";
 
     export default {
 
@@ -38,8 +38,6 @@
 
             relevantColumns: { type: Array, required: true }
         },
-
-        inject: ["missingColumnValues"],
 
         data() {
 
@@ -65,31 +63,40 @@
 
             ...mapGetters([
 
-                "isMissingValue",
                 "valueDescription"
             ]),
 
+            ...mapState([
+
+                "missingColumnValues"
+            ]),
+
             tableItems() {
-                const tmp = [
-                    {
-                        column: 'sex',
-                        description: 'my description',
-                        value: 'my value'
+                let missingValueArray = []
+
+                for ( let column of this.relevantColumns ) {
+                    if ( Object.keys(this.missingColumnValues).includes(column) ) {
+                        for ( let missing_value of this.missingColumnValues[column] ) {
+                            const description = this.valueDescription(column, missing_value);
+                            missingValueArray.push(
+                                {
+                                    column: column,
+                                    description: description === null ? "" : description,
+                                    value: missing_value
+                                }
+                            )
+
+                        }
                     }
-                ]
-                return tmp
+                }
+                return missingValueArray
             }
         },
 
         methods: {
 
             removeColumn(p_tableItem) {
-
-                // 1. Remove the item from table data
-                this.tableItems = this.tableItems.filter(item =>
-                    item.column === p_tableItem.column && item.value !== p_tableItem.value );
-
-                // 2. Remove this value from the column's missing value list in the store
+                // Remove this value from the column's missing value list in the store
                 this.$emit('remove:missingValue', p_tableItem);
             }
         }
--- a/components/annot-missing-values.vue
+++ b/components/annot-missing-values.vue
@@ -1,7 +1,9 @@
 <template>
 
     <div>
-        <b-card class="annotation-card">
+        <b-card
+            no-body
+            class="annotation-card">
 
             <b-card-header>{{ uiText.title }}</b-card-header>
 
--- a/components/annot-missing-values.vue
+++ b/components/annot-missing-values.vue
@@ -74,7 +74,7 @@
             ]),
 
             tableItems() {
-                let missingValueArray = []
+                let missingValueArray = [];
 
                 for ( let column of this.relevantColumns ) {
                     if ( Object.keys(this.missingColumnValues).includes(column) ) {
@@ -86,12 +86,12 @@
                                     description: description === null ? "" : description,
                                     value: missing_value
                                 }
-                            )
+                            );
 
                         }
                     }
                 }
-                return missingValueArray
+                return missingValueArray;
             }
         },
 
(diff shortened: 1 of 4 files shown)


components/annot-missing-values.vue
```diff
@@ -68,133 +65,41 @@
...
-
-            isValidString(p_value) {
-
-                let isValid = true;
-
-                return isValid;
-            },
+        methods: {
 
             removeColumn(p_tableItem) {
```
jarmoza, line 100: Check for trailing whitespace in this function

components/annot-missing-values.vue
```diff
@@ -68,133 +65,41 @@
...
-                    missingValuesLists[columnName] = [];
+            ...mapState([
 
-                    // C. Determine value invalidity based on the data type of this category
-                    for ( const value of this.uniqueValues[columnName] ) {
+                "missingColumnValues"
+            ]),
 
-                        switch ( this.dataType ) {
+            tableItems() {
```
jarmoza, line 76: Remember to add semicolons. A couple places in this function. Not sure why our linter is not picking up on this.

pages/annotation.vue
```diff
@@ -151,6 +151,10 @@
                 pageName: "download",
                 enable: this.isDataAnnotated
             });
+
+            // TODO: remove this, just a temporary way to add any missing values to the store
+            const missingLists = {"sex": [0, 1]}
```
jarmoza, line 156: Missing semicolons here

store/index.js
```diff
@@ -465,12 +465,14 @@ export const mutations = {
     },
 
 	setMissingColumnValues(p_state, p_missingColumnValues) {
+        // This method merges incoming updated missingColumnValues records with the missingColumnValues
+        // object in the store. Because the incoming changes can be incomplete (e.g. only contain updated
+        // records of a single column), we cannot just overwrite the store object with them.
+        // However, because of how reactivity in Vue works, we can also not simply overwrite the affected columns
+        // (i.e. keys) in the object, because that will break reactivity.
+        // The below pattern via assign sovles this problem. See here: https://v2.vuejs.org/v2/guide/reactivity.html
+        p_state.missingColumnValues = Object.assign({}, p_state.missingColumnValues, p_missingColumnValues);
```
jarmoza, line 474: Nicely done!

jarmoza (commented):
Some minor formatting changes. We have to check why our linter has not picked up on missing semicolons.

Feel free to merge after you fix the formatting issues.

components/annot-missing-values.vue
surchs: I checked, it's not set as a rule. Try `npx eslint --print-config components/annot-missing-values.vue`. In fact, `no-extra-semi` is set in the rules. Feel free to add to https://github.com/neurobagel/annotation_tool/issues/125 if you find other things

components/annot-missing-values.vue
surchs: Thanks, but I don't see any trailing whitespace in this method. Do you?

jarmoza: The linter picked up whitespace at lines 193 and 197, yes.

Issue #138, opened 2022-05-09
Every tool group I assign columns to shows up on the annotation page as a tab


surchs:

surchs:
Hey @jarmoza, what would be an ungrouped assessment tool to you? If you mean a tool that's described by only one column, then yes, this should be possible. But I think the way to do that would be to create a tool group that has only one column. There shouldn't be a way to describe an assessment tool column without assigning it to a tool group. Is that what you were asking about?

jarmoza:
> There shouldn't be a way to describe an assessment tool column without assigning it to a tool group. Is that what you were asking about?

Yes, in other words, "Must all assessment tools be grouped?" Can the user decides that they don't want to utilize that functionality?

surchs:
I would say no. As few exceptions as we can get away with. Having a single column tool seems pretty OK to me. Are you thinking of something that would affect the implementation or are you thinking about user experience?

jarmoza:
Okay. So the result of discussion over Slack video is that in order to move onto the Annotation page, all columns that have been categorized as an assessment tool must be part of a group – even if they are a single column.

Issue #118, opened 2022-05-02
I can group assessment tool columns together


surchs:
When I categorize columns, I want to be able to indicate that multiple columns belong to the same assessment tool (e.g. subscales of the same tool), so that
- I don't have to label every column individually with the same concept
- I can indicate the columns that are not allowed to have missing values for this tool to be considered available for a subject

Done when:

- [ ] I can create a new tool group with the user interface on the categorization page
- [ ] A new tool group gets a default name (e.g. tool 1) if I don't provide one
- [ ] Adding columns to tool groups works similarly to mapping columns to age, sex, and diagnosis. I can click on a tool group (or make a new one), and then click on a column to assign it to this tool group
- [ ] I can remove columns from a tool group via the categorization page UI
- [ ] A column can only be added to a single tool group.

jarmoza:
Separate piece of the interface will be for grouping columns, assigning a name - and possibly assigned a ruleset heuristic

surchs:
One thing that would important as well is to document how these groups of assessment tools are stored in the global store so that we can build against this organization when we implement the assessment tool annotation process next.

jarmoza:
> I can indicate the columns that are not allowed to have missing values for this tool to be considered available for a subject

This should be part of a separate issue. I've created one here: https://github.com/neurobagel/annotation_tool/issues/131

PR #150, opened 2022-05-18
Assessment tool group tabs and enhanced functionality

jarmoza:
-[] Assessment tool groups are now placed in the Vuex store and used to generate separate annotation tabs on the annotation page. 
-[] All columns categorized as an 'Assessment Tool' must now be grouped in order to proceed to the annotation page, and will be found within their corresponding tool group tab on the annotation page.
-[] Tool group name and column (tool) list are fully editable as well as removable. In order to begin modifying a tool group, select its row in the tool group table. The tool group's current name as well as all available tools for grouping (its current tools and other un-grouped tools) will show up in the same interface wheretool group creation happens. Button text will change to '~ Modify Tool Group'
-[] Uncategorizing columns in the column linking table will remove them from currently saved tool groups. If all columns in a tool group are uncategorized in this way, the group itself is deleted
-[] Changes made in the tool group interface on the categorization page are reflected on the annotation page (group name and tools per group).

This PR addresses functionality listed in both https://github.com/neurobagel/annotation_tool/issues/118 and https://github.com/neurobagel/annotation_tool/issues/138. See each issue for the desired functionality checklists.


components/annot-tab.vue
```diff
@@ -82,6 +82,12 @@
                 const columnList = [];
                 for ( const columnName in this.columnToCategoryMap ) {
 
+                    // If this tab is an assessment tool group, make sure this column is in its toolset
+                    if ( Object.prototype.hasOwnProperty.call(this.details, "groupName") &&
```
surchs, line 86: The `Object.prototype...` thing made me look twice. If I understand right, ESlint flags this because the default `this.details.hasOwnProperty` could fail if `this.details` happens to have a property called `"hasOwnProperty"`? The recommended way to work around this (if legacy browser support isn't a barrier) [seems to be to use `hasOwn` ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn) instead, which is the replacement of `hasOwnProperty`.

I'd suggest we use `hasOwn` here or if not add a comment why were are using this less expected form.

components/categ-tool-group.vue
```diff
@@ -165,34 +189,147 @@
...
+                deep: true,
+                handler(p_newColumnToCategoryMap, p_oldColumnToCategoryMap) {
+
+                    // Note: `newValue` will be equal to `oldValue` here
+                    // on nested mutations as long as the object itself
+                    // hasn't been replaced.
+
+                    // 1. Check to see if any columns have been unlinked as assessment tools
+                    const toBeRemoved = [];
+                    for ( const item of this.assessmentToolGroups.items ) {
```
surchs, line 259: There is something weird going on here when the `columnToCategoryMap` have been changed (e.g. by removing a column from a tool group in the annotation window). The problem seems to be that although `this.assessmentToolGroups` clearly has a property `items` with a non-empty array, directly accessing this array via `this.assessmentToolGroups.items` (e.g. `console.log(this.assessmentToolGroups.items)` shows an empty array. 

Take a look here:
annot_tool_bug
If you access the variable directly or watch it via vue-dev tools, all looks normal. But obviously you cannot iterate over an empty array so you never get to check if anything has to be removed here.

As I mention in the review, i think the real solution to this is to redesign the way the tool groups are implemented. But we should do this in a later PR, right now we get what we need and this bug is something we can work around.

surchs (approved):
Thanks for the PR @jarmoza, this is working and looks very nice!
It also solves our need to group columns together so we can annotate them together and then define missing values.

That said, I think in a future PR we should redesign the tool-grouping workflow. The main issue is that we now have an additional thing in the store to keep aligned with the columnToCategoryMapper we already have, and this alignment is hard to do, we need watchers, and refresh methods and other things that are easy to break. See for example the bug I described where, if you assign two columns to a tool, and then remove one of the columns via the annotation window, the column is still assigned to the tool, but no longer designated as a "assessment tool" category.

I think at the core, the "map columns to tool groups" workflow is the same as the "map columns to categories" workflow. The only difference is that the category you are assigning columns to (i.e. the tool group) has to be created first. That is, if we find a way of creating new categories for the assessment tool groups, then we don't have to have separate store variables and UI elements to keep columns and tool groups aligned with columns and categories. 

That said, the functionality right now is all we really need to start using the tool (and to add the assessment tool group annotation functionality) and so I suggest we merge this PR with some minor changes (as commented) and think about my above points in a separate issue.

components/annot-tab.vue
jarmoza: Yup. That's right. `Object.prototype.hasOwnProperty.call` is the suggestion to make sure of the correct behavior. `hasOwn` is newer, considers a few more edge cases, and "is more intuitive." Sounds right.

components/annot-missing-values.vue
```diff
@@ -98,7 +98,12 @@
         methods: {
 
             removeColumn(p_tableItem) {
-                // Remove this value from the column's missing value list in the store
+
+                // 1. Remove the item from table data
+                this.tableItems = this.tableItems.filter(item =>
```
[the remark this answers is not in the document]
jarmoza: I am too a bit puzzled by this line. I don't quite recall writing it. Just looking at what the function is supposed to be doing though, perhaps an answer to this question would clear things up.

When a value is removed from the missing value table how else is `tableItems` being updated? I don't see any other place it's touched other than in `created()`.

components/categ-tool-group.vue
jarmoza: 1. Yes, to redesign. One of the reasons I had wanted to take a look at a potential object-oriented reworking is because things are beginning to feel to disparate/unwieldy in the store and in turn too constrictive (a.k.a. 'brittle'). Too many sources of truth. That said, that's not for now. So let's discuss this smaller internals redesign next.
2. I'm going to take a look into this for a little bit so I can better understand what is happening.

jarmoza: So, from what I can tell there is some disconnect in how you're viewing the object via the console. Debug messages from source reveal `assessmentToolGroups.items` to be correctly populated.

That said, I have produced a fix for the issue of the table not being updated and though we should readdress how this is done in the redesign we mentioned, we have some nice "magic" at hand.

In the annotation pages `unlinkColumnFromCategory` method - which is called when a column is removed from the `annot-column` component, I have added the following third step to appropriately update the tool groups in the store.

```
                // 3. If this column was a grouped tool, remove it from its group
                if ( this.isToolGrouped(p_event.removedColumn) ) {

                    // A. Remove the tool from its group
                    const groupName = this.getGroupOfTool(p_event.removedColumn);
                    this.$store.dispatch("removeToolFromGroup", {
                        group: groupName,
                        tool: p_event.removedColumn
                    });

                    // B. If its former group is now empty of tools, remove the group itself
                    for ( const groupName in this.toolGroups ) {
                        if ( 0 === this.toolGroups[groupName].length ) {

                            this.$store.dispatch("removeToolGroup", { name: groupName });
                        }
                    }
                }
```

This duplicates behavior from the `categ-tool-group` component itself, but applies the changes we want.  Changes are reflected on the categorization page.

And here's the Vue "magic" part. If every tool is removed from its tab on the annotation page (`annot-columns` component), the tab itself is now removed.

PR #156, opened 2022-05-30
Precommit hooks via husky

jarmoza:
Adding pre-commit linting via [husky](https://typicode.github.io/husky/#/).

Though most of the changes after merging master into this branch are (appropriately) linter spacing changes, here are the relevant changes for this PR:

1. .husky/ - This folder contains `pre-commit` the husky config file to run the `lint` command as defined in a `package.json`
2. `package.json` - Contains the pre-commit hook configuration for husky

Likely this PR is a candidate for 'squash and merge'


package.json
```diff
@@ -7,7 +7,13 @@
     "build": "nuxt build",
     "start": "nuxt start",
     "generate": "nuxt generate",
-    "lint": "npx eslint --ext .vue  --ext .js ."
+    "lint": "npx eslint --ext .vue  --ext .js .",
+    "prepare": "husky install"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "npm run lint"
```
surchs, line 15: Between here and `.husky/pre-commit` what is the right place to define the actions that should happen when the hook fires? Does one call the other? Or are both needed. 

I'm thinking of a time in the future when we want to, say, also run some tests on git-commit. Would we add this here to the hooks or in `.husky/pre-commit`?

surchs (approved):
Thanks a lot @jarmoza for getting to this! Very excited to have these pre-commit hooks in. Hopefully that's the last we'll see of the pesky white-space stuff. 

I checked it out locally and confirmed it works (i.e. keeps me from committing stuff with whitespaces) after I do `npm install`. If I understand correctly, running `npm install` is required for this to work (makes sense). I would still say: let's add a line to the README on how to use pre-commit (and maybe why we have it).

Otherwise, good to go! :woman_dancing:

surchs:
Also agreed on squash merge for this guy!

package.json
jarmoza: Good question. In fact, after rereading about husky just now (https://blog.typicode.com/husky-git-hooks-javascript-config/) it turns out that this section of `package.json` is superfluous. As husky is installed via npm it redirects git to look at `.husky` for hooks.

Commands like `lint` can be defined in package.json or they can simply be put into `pre-commit` by hand or automatically like so on the command line: 

`npx husky add .husky/<hook-filename> "<command>"`

In your example it would be `npx husky add .husky/pre-commit "npm test"` and you would define `test` in the `scripts` definition of `package.json`

I will remove the `husky` section of `package.json` here, commit, and merge.

surchs: Ah, interesting. Looks like the `package.json` way was the old way and they have since then changed to the `.husky/` way

jarmoza: Edited the above comment for clarity.

Issue #162, opened 2022-06-03
Create JSON output of annotated input data


surchs:
When I click on the download button on the download page, I want to get a JSON file that contains all of the annotated information for each subject, so that I can go and feed it to a graph store.

- [ ] The JSON file should look something like what we discussed in #7 and be focused on individual subjects (i.e. have subject-type objects high up in the JSON tree).
- [ ] the file shouldn't include data from the original input data that hasn't been annotated (e.g. extra columns)
- [ ] each subject record should only contain data that hasn't been labeled as missing. For example, if a subject had in the raw input an age-column value of "n/a" and this was labeled as missing, then the JSON file should not provide the "age" property for this subject (rather than just leaving the value empty or having "missing").

surchs:
@jarmoza, this is a first draft of the high level things this JSON guy is supposed to do for us. Let me know if any of this doesn't make sense or needs more discussion

PR #163, opened 2022-06-06
Annotation tool now outputs JSON files reflecting the annotated table

jarmoza:
Primary changes:

- Upon clicking the 'download' button on the Download page produces a new subject-centric JSON format of the data in the annotated table (as described in #7 and #162)
- The new JSON is then downloaded to the user's browser downloads folder with the filename: 
`<original data table tsv filename>_annotated_<current timestamp>.json`

Other related changes:

- Filenames of data table and data dictionary files are stored in the store after upload on the landing page
- Fix for getter `getColumnOfCategory` --> `getColumnsOfCategory`

### DIFF
--- a/components/annot-tab.vue
+++ b/components/annot-tab.vue
@@ -71,7 +71,7 @@
 
             ...mapGetters([
 
-                "getColumnOfCategory"
+                "getColumnsOfCategory"
             ]),
 
             filteredDataTable() {
@@ -161,7 +161,8 @@
             this.category = this.details.category;
 
             // 2. Get column marked as subject ID
-            this.idField = this.getColumnOfCategory("Subject ID");
+            // NOTE: Subject ID is only allowed one column
+            this.idField = this.getColumnsOfCategory("Subject ID")[0];
         }
     };
 
--- a/pages/download.vue
+++ b/pages/download.vue
@@ -78,7 +78,6 @@
 
             ...mapGetters([
 
-                "getColumnsOfCategory",
                 "isDataAnnotated"
             ]),
 
--- a/pages/download.vue
+++ b/pages/download.vue
@@ -26,7 +26,8 @@
                 <b-button
                     class="float-right"
                     :disabled="!isDataAnnotated"
-                    :variant="downloadButtonColor">
+                    :variant="downloadButtonColor"
+                    @click="downloadAnnotatedData">
                     {{ uiText.downloadButton }}
                 </b-button>
             </b-col>
@@ -43,6 +44,9 @@
     import { mapState } from "vuex";
     import { mapGetters } from "vuex";
 
+    // Saves file to user's computer
+    import { saveAs } from "file-saver";
+
     export default {
 
         name: "DownloadPage",
@@ -50,6 +54,10 @@
         data() {
 
             return {
+
+                categoryToColumnMap: {},
+
+                jsonSpacing: 4,
 
                 // Size of the file display textboxes
                 textArea: {
@@ -63,7 +71,6 @@
 
                     downloadButton: "Download Annotated Data"
                 }
-
             };
         },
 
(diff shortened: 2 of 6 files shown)


components/annot-tab.vue
```diff
@@ -161,7 +161,8 @@
             this.category = this.details.category;
 
             // 2. Get column marked as subject ID
-            this.idField = this.getColumnOfCategory("Subject ID");
+            // NOTE: Subject ID is only allowed one column
```
surchs, line 164: Maybe this would be clearer here:
```suggestion
            // NOTE: the "Subject ID" category is only allowed to be applied to one column
```

pages/download.vue
```diff
@@ -76,9 +83,23 @@
...
+
+                // Dataset name is original data table filename with no extension
+                return this.dataTable.filename.split(".").slice(0, -1).join(".");
+            },
+
+            defaultOutputFilename() {
+
+                return `${this.datasetName}_annotated_${Date.now()}.json`;
+            },
+
```
surchs, line 102: Just wanted to say: very nice and clean stuff going on here! :tada:

pages/download.vue
```diff
@@ -88,32 +109,103 @@
...
+                const jsonData = this.transformAnnotatedTableToJSON();
+
+                // 2. Open file dialog to prompt the user to name it and
+                // download it to their location of choice
+                this.fileSaverSaveAs(jsonData);
+            },
+
+            fileSaverSaveAs(p_jsonData) {
+
+                var blob = new Blob([JSON.stringify(p_jsonData, null, this.jsonSpacing)], {type: "text/plain;charset=utf-8"});
```
surchs, line 133:
```suggestion
                const blob = new Blob([JSON.stringify(p_jsonData, null, this.jsonSpacing)], {type: "text/plain;charset=utf-8"});
```

pages/download.vue
```diff
@@ -88,32 +109,103 @@
...
+                saveAs(blob, this.defaultOutputFilename);
+            },
+
+            refreshCategoryToColumnMap() {
```
surchs, line 137: I'm pretty sure this is a duplicate from elsewhere in the app, no?
Let's at least add a `// TODO` flag or something like it so we don't forget to refactor this!

pages/download.vue
```diff
@@ -88,32 +109,103 @@
...
+
+            transformAnnotatedRowToSubjectJSON(p_row) {
+
+                // 0. Determine all columns of the original data that have been categorized
+                const availableCategories = Object.values(this.columnToCategoryMap);
+
+                // 1. Create the new subject object, key by key
+                const subjectJSON = {};
+
+                // NOTE: Currently only one column is allowed for subject ID, age, and sex
```
surchs, line 171:
```suggestion
                // NOTE: Currently the categories subject ID, age, and sex must be applied to only one column each
```

surchs (approved):
Yay :tada:. Awesome milestone @jarmoza! We definitely need to celebrate this!
I think this is something you should consider demonstrating to the group on Thursday. Then we can also see how this sprint-release demo thing works for us.

To the PR: tried it out, confirmed it works! I think there is a bit of code duplication going on in the download.vue page but that's alright for now. Would still suggest to add a flag so we'll find it again during refactor!

Otherwise all looks great to me, let's go ahead and merge this!

pages/download.vue
jarmoza: Danke!

pages/download.vue
jarmoza: Oops. Good catch. Probably copy-pasted from the example.

pages/download.vue
jarmoza: No, this is in fact a method to create/refresh a reverse of the column to category map in the store. `categoryToColumnMap` is a local object only used by the download page. Made sense to update it in some way in case a user altered the `columnToCategoryMap` in the store.

PR #165, opened 2022-08-04
First Cypress End to End Testing Implementation

jarmoza:
This PR is being set up for discussion of our first implementation of end to end testing using Cypress.

Comments on commits thus are added below.


cypress.config.js
```diff
@@ -0,0 +1,23 @@
+const { defineConfig } = require("cypress");
+
+module.exports = defineConfig({
+
+    e2e: {
+
+        baseUrl: "http://localhost:3000",
```
jarmoza, line 9: `baseUrl` is used by calls to  `cy.visit`

package.json
```diff
@@ -9,7 +9,8 @@
     "generate": "nuxt generate",
     "lint": "npx eslint --ext .vue  --ext .js .",
     "prepare": "husky install",
-    "heroku-postbuild": "npm run build"
+    "heroku-postbuild": "npm run build",
+    "cypress:open": "cypress open"
```
jarmoza, line 13: As per suggestion here: https://docs.cypress.io/guides/getting-started/opening-the-app

Cypress can be started via `npm run cypress:open` instead of writing the path to cypress in npm's `bin`

cypress/support/commands.js
```diff
@@ -22,4 +22,94 @@
 //
 //
 // -- This will overwrite an existing command --
-// Cypress.Commands.overwrite('visit', (originalFn, url, options) => { ... })
\ No newline at end of file
+// Cypress.Commands.overwrite('visit', (originalFn, url, options) => { ... })
+
```
jarmoza, line 26: `nextPageByButton`, `nextPageByNav`, and `categorizeColumn` are implemented for clarity in commonly reused test code.

`passIndex` and `passCategorization` remain in use for UI-based (not programmatic state-based) testing.

components/tool-navbar.vue
```diff
@@ -22,6 +22,7 @@
                     <b-nav-item
                         v-for="(navItem, _key) in navItems"
                         :active="pageName === navItem.fullName"
+                        :data-cy="'menu-item-' + navItem.pageName"
```
jarmoza, line 25: As per recommended best practice, any attribute accessed as an identifier by cypress should be separate from identifiers used by the app itself in order to avoid brittleness.

Our current standard is `data-cy` for the attribute name.

cypress/e2e/index/select-bothfiles.cy.js
```diff
@@ -0,0 +1,43 @@
+// Description: User selects data table AND data dictionary
+// Expected results: Data dictionary selection does not enable next page nav, but data table selection does
+
```
jarmoza: The testing functionality in this and the following test files in this commit (each `describe` + `it` function definitions) are the first implementation of UI-based testing for the index page before later test organization (`describe`, `context`, and multiple `it` functions per context) was established.

cypress/e2e/categorization/bothtypes-categorization.cy.js
```diff
@@ -0,0 +1,77 @@
+// Description of task: user selects non- and tool group categories, paints column with tool group paint,
+// creates tool group label, groups columns, paints columns with non-tool group categories
+// Expected results: Annotation nav and Next page button are enabled
+
+// Import globals
+// import "./e2e";
+
+describe("bothtypes-categorization", () => {
```
jarmoza: The testing functionality in this and the following test files in this commit (each describe + it function definitions) are the first implementation of UI-based testing for the categorization page before later test organization (describe, context, and multiple it functions per context) was established.

cypress.config.js
```diff
@@ -1,5 +1,7 @@
 const { defineConfig } = require("cypress");
 
+const { isFileExist, findFiles } = require('cy-verify-downloads');
```
jarmoza: This plugin is used for verifying json file download on the download page.

cypress.config.js
```diff
@@ -11,8 +13,13 @@ module.exports = defineConfig({
       }
     },
 
+    downloadsFolder: "cypress/downloads",
```
jarmoza: Explicit mention of the downloads folder location where test will put the downloaded json file from the download page.

cypress/e2e/download/download-file.cy.js
```diff
@@ -0,0 +1,23 @@
+// Description of task: User clicks on the 'download file' button
+// Expected results: File is verified as downloaded
+
+describe("download-file", () => {
```
jarmoza: The testing functionality in this and the following test files in this commit (each describe + it function definitions) are the first implementation of UI-based testing for the download page before later test organization (describe, context, and multiple it functions per context) was established.

cypress/support/commands.js
```diff
@@ -101,6 +101,26 @@ Cypress.Commands.add("passCategorization", () => {
     cy.nextPageByButton();
 });
 
+// Go through annotation page, saving default age annotation
+Cypress.Commands.add("passAnnotation", () => {
```
jarmoza: `passAnnotation` remains in use for UI-based (not programmatic state-based) testing.

.github/workflows/main.yml
```diff
@@ -0,0 +1,17 @@
+name: Cypress Tests
```
jarmoza, line 1: This GH action script will be amended as necessary in a coming commit.

cypress.config.js
```diff
@@ -1,34 +1,43 @@
...
+            devServer: {
+
+                bundler: "webpack",
+                framework: "nuxt"
+            }
+        },
+
+        // Downloads location and behavior
+        downloadsFolder: "cypress/downloads",
+        trashAssetsBeforeRuns: true,
```
jarmoza, line 22: Every time a test runs, files (e.g. downloaded files from a download page test) from previous tests are deleted.

cypress/support/commands.js
```diff
@@ -24,6 +24,58 @@
 // -- This will overwrite an existing command --
 // Cypress.Commands.overwrite('visit', (originalFn, url, options) => { ... })
 
+// Calls action in the Nuxt store
+Cypress.Commands.add("dispatchToNuxtStore", (p_action, p_data) => {
```
jarmoza, line 28: `dispatchToNuxtStore`,`loadDataDictionary`, and `loadDataTable` are implemented for clarity in commonly reused test code.

cypress/fixtures/tests/categorization/default-test-config.json
```diff
@@ -0,0 +1,5 @@
+{
+    "source_folder": "examples/good/",
```
jarmoza, line 2: A set of variant test configuration files will need to be created for the initial form of our overall test 'suite' for this app (and others).

This can be composed of the various pairings that @surchs created for our initial manual testing of the annotation tool.

cypress/e2e/page/categorization-pagetests.cy.js
```diff
@@ -0,0 +1,292 @@
+describe("tests on categorization page via programmatic state loading", () => {
```
jarmoza: Current organization for tests is as follows:

page-based tests found in `cypress/e2e/page` - these are tests in which page state is produced programmatically and a series of tests on the page's functionality are initiated

ui-based tests found in `cypress/e2e/ui` - these are tests where pages are successively interacted with and navigated via ui interaction

Test files themselves also have the following _internal_ organization:

1. `describe` - a containing function given the title of a summary description of this file's tests
with a `beforeEach` that contains code run before each testing context listed within
2. `context` - one or more specific testing contexts (within that overall purpose of tests in this file) given a title describing that testing context. (In this case, the standard is that each context represents a different set of input data files.) Each context contains a `beforeEach` run before each individual test (`it`) run in this testing context
3. `it` - An individual test within that testing context

cypress/support/commands.js
```diff
@@ -93,7 +93,7 @@ Cypress.Commands.add("nextPageByNav", (p_navItemName) => {
 });
 
 // Go through index page and select participants tsv and json dictionary files, advance to categorization page
-Cypress.Commands.add("passIndex", () => {
+Cypress.Commands.add("passIndex", (p_dataTableFilepath, p_dataDictionaryFilepath) => {
```
jarmoza, line 96: These `pass` commands now take variable input paths from the current testing context.

cypress/e2e/ui/categorization-uitests.cy.js
```diff
@@ -0,0 +1,239 @@
+describe("tests on the categorization page via ui interaction", () => {
```
jarmoza, line 1: UI-based tests now follow the same organization used by page-based/programmatic-state-based testing.

cypress/.eslintrc.json
```diff
@@ -5,7 +5,7 @@
 
     "rules": {
         // "cypress/no-assigning-return-values": "error",
-        "cypress/no-unnecessary-waiting": 0
+        // "cypress/no-unnecessary-waiting": "error",
```
jarmoza, line 8: No specific cypress linting rules are used, just the recommended linting rules. Removing this line to ensure any further `no-unnecessary-waiting` errors are not suppressed.

surchs (changes requested):
Thanks @jarmoza for setting up all of these tests. Very excited that we have a test suite now and can keep expanding it! This is a super important step for us to add new features and have confidence in the annotator. Very cool!

I have made some comments in the code, please take a look or let me know what you think.

Two more general comments:

1 - I don't understand the distinction between page and ui
(I saw your comment just commenting here to keep things together). Both are e2e tests. We always want to test the UI, and we never want to test the programmatic state setting. The only difference to me is that we want to test a route through the UI only once. And once we know it works, we can take the programmatic shortcut for any subsequent tests. So for example: we only test the select-and-upload functionality on the `index.vue` page once, and for subsequent tests we "upload" the file not via the UI but programmatically.

All that is to say: let's merge `page` and `ui` and make sure that each UI route is tested at least once, and then used programmatically afterwards.

2 - Let's keep it simple as long as we can
We're not using many different example data files yet, so I don't think we need the added complexity of fixtures and aliases in the `beforeEach` sections of the tests (I made separate inline comments about these). We're also not using component testing yet, so let's remove the config files for that to keep things clean. We can always re-generate them once we need them. And lastly, this is a big PR - I understand this is a big piece of work. But let's try to aim for smaller, more incremental changes in the future. They are easier to review and and we can get faster feedback and discuss things. It also makes for more logical PRs, because if it's open for a long time, you can get old solutions from earlier in the PR that live on in some commented out piece of code and so on. We can always add more stuff later if we need it!

.eslintrc.js
```diff
@@ -19,7 +20,11 @@ module.exports = {
         $nuxt: true,
 
         // 'structuredClone' is available as of ES6 (ca. 2015), but ESLint does not yet know about it
-        structuredClone: true
+        structuredClone: true,
+
+        // Cypress
```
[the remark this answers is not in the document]
jarmoza: Good catch. `require` and `module` do not appear to be necessary. My guess is this was part of one of the suggested via stackoverflow means of integrating Nuxt and Cypress.

Removing them.

components/categ-tool-group.vue
```diff
@@ -33,6 +34,7 @@
                 <b-input-group>
                     <label for="column-multiselect">Columns:&nbsp;</label>
                     <b-form-select
+                        data-cy="column-multiselect"
```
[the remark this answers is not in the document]
jarmoza: Do you mean for the value of the `data-cy` attribute?

components/tool-navbar.vue
jarmoza: Yes, I considered whether to select by ID value or UI label too – if I'm understanding your comment correctly.

Whatever we do should be standard in our testing implementation, but my intuition is not to rely on the label text but rather than the html attribute value. Arguably, it is less likely for the attribute value to change.

cypress.config.js
```diff
@@ -0,0 +1,43 @@
...
+
+const { isFileExist, findFiles } = require("cy-verify-downloads");
+
+module.exports = defineConfig({
+
+    e2e: {
+
+        baseUrl: "http://localhost:3000",
+
+        component: {
```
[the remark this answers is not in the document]
jarmoza: Yup. Good catch. And I'm guessing it didn't error because Cypress would not be looking for a `component` property in the `e2e` definition.

cypress/.eslintrc.json
jarmoza: Technically no need anymore. I was experimenting with them earlier on. I can just delete the specific `rules` definition entirely now.

cypress/e2e/page/categorization-pagetests.cy.js
```diff
@@ -0,0 +1,292 @@
+describe("tests on categorization page via programmatic state loading and store interaction", () => {
+
+    beforeEach(() => {
+
+        // Standard setup for annotation tool testing
+
+        // A. Set viewport size
+        // NOTE: Standard 13-inch laptop screen to start, but this can be expanded
+        cy.viewport("macbook-13");
```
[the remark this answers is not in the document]
jarmoza: Agreed. Good suggestion!

components/categ-tool-group.vue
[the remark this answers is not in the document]
surchs: yeah, exactly. because we can chose any name here we can use one that is descriptive and makes for easy to read tests. if we end up liking that name, we could consider updating the id name eventually as well.

cypress/e2e/page/categorization-pagetests.cy.js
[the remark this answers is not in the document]
jarmoza: Secondary comment here. I'm creating an `appSetup` command, but in the future I'm guessing we will likely want to parametrize it for different test setups (i.e. testing across different viewport sizes).

cypress/e2e/page/categorization-pagetests.cy.js
```diff
@@ -0,0 +1,292 @@
...
+
+            // Action 2
+
+            // Link the first column in the column select table with the subject ID category
+            cy.get("[data-cy='column-linking-table'] tbody tr td")
+                .contains("participant_id")
+                .click();
+
+            // Assert 2 - Nav and next button should be enabled
+            // Assert 1
```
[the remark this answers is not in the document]
jarmoza: I'm following the testing flow of Setup -> Action(s) -> Related assert(s). The commenting scheme I came up with to organize this with a description for the overall action and related asserts. The numbering is to keep track of the nth action/asserts of this individual test.

cypress/e2e/page/categorization-pagetests.cy.js
```diff
@@ -0,0 +1,292 @@
...
+            cy.get("[data-cy='menu-item-annotation'] a")
+                .should("have.class", "disabled");
+            cy.get("[data-cy='button-nextpage']")
+                .should("have.class", "disabled");
+
+            // Action 2
+
+            // Link the first column in the column select table with the 'Subject ID' category
+            cy.get("[data-cy='column-linking-table'] tbody tr td")
+                .contains("participant_id")
```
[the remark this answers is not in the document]
jarmoza: Looking into this turned out to be more interesting than I realized it would be.

Firstly, agreed about the genericization re: the specific string. And for the simple categorization test that's just fine – with one exception (see below). In fact, that's how I first implemented this and other table row selects. You call `find("tr")` and use the `eq` function to select the nth index. 
What I've noticed is that sometimes `eq` can select an index you don't expect. I noticed this on the linking table so I just switched all table row selects to use the select by text method. There is likely a reason for this unexpected behavior. Either it's something we're doing (i.e. not specifying a row index correctly) or something with Vue/Cypress mixed.

The "interesting" part is that your pointing this surfaces something we haven't smoothed out since we were developing an app prototype and not something with more generic logic. Within the categorization pages `nextPageAccessible` we have the following code:

```
                // 4. Make sure one (and only one) column has been categorized as 'Subject ID'
                let subjectIDFound = 0;
                for ( const column in this.columnToCategoryMap ) {
                    if ( "Subject ID" === this.columnToCategoryMap[column] ) {
                        subjectIDFound += 1;
                    }
                }
                const singleSubjectIDColumn = ( 1 === subjectIDFound );
```

Obviously this code will need to be genericized for different formats, as you mention. So 'simple' categorization requires at least subject ID to be selected.

For now it makes sense to keep the text select specifically for 'Subject ID', but I have added a new issue for this future task here: https://github.com/neurobagel/annotation_tool/issues/167

cypress/e2e/page/categorization-pagetests.cy.js
```diff
@@ -0,0 +1,292 @@
...
+            // Action 3
+
+            // Select 'Diagnosis' category
+            cy.get("[data-cy='categorization-table']")
+                .contains("Diagnosis")
+                .click();
+
+            // Link the second column in the column select table with the 'Diagnosis' category
+            cy.get("[data-cy='column-linking-table'] tbody tr td")
+                .contains("group")
```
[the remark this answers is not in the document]
jarmoza: Sure on this one. As mentioned above, 'Subject ID' is currently a hardcoded category we need to be selecting for passing criteria.

I will adjust the table selects to be more generic with the exception of 'Subject ID' (for now).

cypress/e2e/page/index-pagetests.cy.js
```diff
@@ -0,0 +1,167 @@
+describe("tests on the index page via store interaction", () => {
```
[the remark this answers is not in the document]
jarmoza: @surchs Can you clarify? I've currently separated out programmatic state tests by page. In this case, we have the rather simple index page where state is loaded and we test to see if that state load results in the desired access to the categorization page.

I could see removing this file entirely though and just relying on the UI e2e test entirely for this page. Is that what you mean?

cypress/fixtures/tests/categorization/default-test-config.json
jarmoza: Totally, re: separate PR.

cypress/support/commands.js
jarmoza: That's correct. The other reason I'm mentioning these functions - and I'll continue this thought in your other general comment on UI-based testing - is that these pass functions may be rendered moot depending on how we decide to configure the UI e2e test(s).

More on that in your general comment on the PR.

jarmoza:
@surchs I think this PR is almost fully baked. I will follow up with commit(s) based on your comments and request review.

Just wanted to make some general comments re: our PR process and the UI-based testing for upcoming PR(s)

1. Regarding the longer PR, I think this is a good chance to discuss how we would like initial (large) features to be added to our repos for ourselves and future collaborators. Inevitably, an initial PR for a feature will be larger than successive, iterative PRs for the same feature (e.g. we don't want to introduce a half baked feature to `master`). In the case of this PR, I think there were several logical chunks that could be split up (though maybe not obvious at first): 1. Cypress initial config 2. Initial test implementation + GH action config 3. Next iteration of test implementation 4. Further config, and so forth... I think the best way to handle this is to identify the chunks of work the dev sees the feature requiring and to plan for the first PR to be a workable feature and then to follow up on the other chunks in successive PRs.

2. Regarding the UI based testing, there are several ways we can go with this and my thought on the implementation has changed the more I've worked with Cypress. Right now we have programmatic state-based tests for pages and I think this makes sense at least for the categorization, annotation, and download pages (and maybe the index page). And then I've separated out UI tests by page utilizing the `pass` commands. However, perhaps it makes more sense to design UI tests as true e2e tests with separate e2e test files differing by the path through the UI and/or input data?

cypress/e2e/page/categorization-pagetests.cy.js
```diff
@@ -0,0 +1,292 @@
...
+            // Assert 1
+            cy.get("[data-cy='menu-item-annotation'] a")
+                .should("not.have.class", "disabled");
+            cy.get("[data-cy='button-nextpage']")
+                .should("not.have.class", "disabled");
+        });
+
+        // Description of task: User selects multiple non-tool group category and paints multiple columns
+        // Expected results: Annotation nav and Next page button are enabled
+        it("multiple categorization", () => {
```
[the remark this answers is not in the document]
jarmoza: Yes, the `it` descriptions can be more specific.

jarmoza:
For some reason I can't find the comment on moving setup code to the support file `e2e.js`, but just to note that `beforeEach` in each describe had to go back in place for app setup. Unfortunately we cannot issue `cy` commands (i.e. `viewport` outside of test files.

cypress/e2e/page/categorization-pagetests.cy.js
```diff
@@ -0,0 +1,292 @@
...
+
+        // A. Set viewport size
+        // NOTE: Standard 13-inch laptop screen to start, but this can be expanded
+        cy.viewport("macbook-13");
+    });
+
+    // Categorization page tests with 'good' data files
+    context("categorization with good data", () => {
+
+        beforeEach(() => {
```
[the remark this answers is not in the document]
jarmoza: While my predilection as an engineer is to build towards the immediate future, this makes sense for now. Also, though I have done some promise-based coding in the past, I am still unfamiliar with the typical best practices of code organization for nesting functions based on them.

All that said, I like this implementation much better and it can be easily parametrized in the near future.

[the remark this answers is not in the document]
jarmoza: This is now in command function `loadTestDataIntoStore`. This implementation is currently only being used in the `categorization-pagetests.cy.js` file, however. Since the only action occurring in the `index-pagetests.cy.js` file is loading the data table and/or dictionary into the store and asserting if navigation to the categorization page has been enabled, it doesn't make sense to put the above implementation there.

(This also points toward the lack of a need for the index pagetest in the first place.)

jarmoza:
@surchs Okay. Good to go now.

cypress/e2e/page/index-pagetests.cy.js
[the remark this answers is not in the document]
surchs: Yes, exactly. We want to make sure that the path a user can take here works as expected. So we need to test that "click button, select file, upload file" has the expected consequence of "now a button is no longer disabled". A user wouldn't directly interact programmatically with the store, so we don't need to test whether a programmatic store manipulation would have the expected effect on the UI.

The tests we want for the index page are already implemented in `index-uitests.cy.js`, so I agree with removing this file.

cypress/e2e/page/categorization-pagetests.cy.js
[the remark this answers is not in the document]
surchs: I see. But here you have both `Assert 2` and `Assert 1`, and I am not sure what that refers to. I like the `Assert` comments where the reason behind the expectation isn't obvious from the code. E.g. `Nav and next button should not yet be enabled` is helpful, because I don't know what the app should do after I click on "Subject ID". I would even consider adding: `" because no column has been categorized yet"` or something like that.

On the other hand, I don't find the comments about actions helpful, because they restate what follows in the code (e.g. `"Select subject ID category"`). So I would consider getting rid of them

cypress/e2e/page/categorization-pagetests.cy.js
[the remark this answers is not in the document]
surchs: Thanks for the issue. Just to be clear: my comments refer to the `[data-cy='column-linking-table']` selector, aka the thing on the right. There I suggest not selecting based on text because that's brittle. We expect there to be a column called `"participant_id"`, but what if we end up using an example that calls this `"subject_number"`. We wouldn't want the test to fail because of that. 

So my point is: let's just pick any column on the right (e.g. the first one) because at this point our app doesn't care if we picked the right one as long as one is picked.

cypress/e2e/page/categorization-pagetests.cy.js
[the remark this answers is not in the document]
surchs: Same as above: I'm not referring to `[data-cy='categorization-table']` here where I think text-based selection makes sense (after all, we control the category names). I'm talking about the `[data-cy='column-linking-table']` selector (for table on the right)

cypress.config.js
```diff
@@ -0,0 +1,35 @@
...
+        setupNodeEvents(on, config) {
+
+            // Implement node event listeners here
+
+            // cy-verify-downloads plugin
+            on("task", { isFileExist, findFiles });
+        }
+    },
+
+    component: {
```
surchs, line 25: Not sure if I missed this section the first time. It's in the right place. But since we aren't using the component framework, I think we should add that when we start using it and remove it for now.

surchs (changes requested):
Thanks for the updates @jarmoza and the changes!

I'll reply to [your comment](https://github.com/neurobagel/annotation_tool/pull/165#issuecomment-1210860854) in here quickly:

> 1. Regarding the longer PR, I think this is a good chance to discuss how we would like initial (large) features to be added to our repos for ourselves and future collaborators

Agreed, let's discuss this today in person. That's a better format for it.

> 2. Regarding the UI based testing, there are several ways we can go with this and my thought on the implementation has changed the more I've worked with Cypress.

Totally understandable. As they are right now, I think the distinction of `page` and `ui` tests is not needed and should be combined. The goal is to test paths through the app from a users perspective, so we should test the UI. And because the later pages depend on state set by previous ones, it also makes sense to not go through the UI route to set this stage every time. But we also don't need to test the programmatic state setting itself. 

I think the strategy here is to test every route through the app via the UI at least once. And then use programmatic state setting for subsequent tests that build on it. So for example: test that we can click the upload button on the index page, select a file, have it upload and validate ... all via the UI once. And then when we test the categorization page, set the store state for this test (i.e. the dataTable and dataDictionary) programmatically, but test the categorization page again via the UI (i.e. clicking on categories and column names ...).

So I'd say:
1. let's keep `index-uitests.cy.js` because it is a UI based test of the index page
2. remove `index-pagetest.cy.js` because it is a direct copy of 1. that test programmatic store changes.
3. let's keep `categorization-pagetests.cy.js` because it is a UI test of the categorization page that handles the index-page based state setting programmatically and not via the UI
4. remove `categorization-uitests.cy.js` because it is a direct copy of 2 that sets initial state by re-testing the UI route through the index page (although that's abstracted away into a separate function).

With that, let's remove all of the `pass_...` functions from `cypress/support/commands.js` and refactor the tests themselves to live under `cypress/e2e` and drop the `-pagetests` or `-uitests` suffix.

Take a look at any of the comments that are still open and then with these changes, I think this is good to go! :tada: we'll have our test suite to build out!

jarmoza:
Last commit is up for this PR. I've done a rework of the files including the requested modifications from @surchs with some additions.

The new organization in the `e2e` folder will include "app" and "page" tests. The `app` subfolder will include true end to end tests of the annotation tool and the `page` folder will include test files per page that goes through and tests out all UI on each page. With the exception of the index page, all other pages will have their state programmatically loaded. (Annotation and Download page test files are still to be implemented/committed).

The second addition is a place for future dynamic testing using different datasets. I've retained the ability to load different data sets through a json config file, but am now using the recommend way of doing such tests as seen here (https://github.com/cypress-io/cypress-example-recipes/tree/master/examples/fundamentals__dynamic-tests), and more specifically this example (https://github.com/cypress-io/cypress-example-recipes/blob/master/examples/fundamentals__dynamic-tests/cypress/e2e/fixture-spec.cy.js). A list of dataset config json are added in an array via `require` and a `forEach` iterates through the list.

Included in the cleanup aside from standardization of testing and commenting are some helper functions in `commands.js` that are called for widely reused functionality like column categorization and next page nav/button status assert.

jarmoza:
One additional note: There is currently a bug with the `verifyDownload` functionality at the end of the simple e2e test. I have created an issue for this here: https://github.com/neurobagel/annotation_tool/issues/172.

surchs:
Hey @jarmoza. Looks good to me! Congrats on this implementation, that's a big step :tada:!

PR #168, opened 2022-08-15
Remove the old mockup image

### LINKED ISSUE #133: Remove the old mockup image
We still have the screenshot of the annotation page mockup inside the repo under `/assets`. This should be removed because we don't need it anymore. 

The only other file in `/assets` is the `main.css` and this is in a subdir `/assets/css`, we should consider to:
- pull the stylesheet up one level to `assets`
- rename the `css` dir into `style`
- do nothing and have an emtpy repo

surchs:
Image is no longer needed. Was used for demo only

Fixes #133


jarmoza:
Obviously, you're all clear to merge this @surchs.

jarmoza (approved):

PR #191, opened 2022-09-23
Update output format to data model

### LINKED ISSUE #189: Change the annotation output format
We now have a type of data model for our meta-data: https://github.com/neurobagel/bagelbids/blob/main/bagelbids/models.py

The format that the annotator outputs was just a first guess at what this model should eventually look like. The problem is that we extract the BIDS data based on this model, and now that the annotator output doesn't match it, merging the annotator-output and the parsed BIDS data is getting more tricky.

So I suggest we change the output of the annotator so that it aligns with this form:

```
 {
            "identifier": "sub-123456",
            "age": 1,
            "sex": "male",
            "diagnosis": [
              {
              "identifier": "http://snomed.org/somecode1234
              }
            ],
            "assessment_tool": [
              {
                "identifier": "http://cognitiveatlas.com/somecode1234"
              }
            ]
        },
```

The main changes are:
1. `id` -> `identifier`
2. diagnosis and assessment_tool arrays contain objects, not strings

surchs:
This changes the output format to the current data model.

This should have been tested as well, however as explained in the issue, writing a test for this is going to be pretty challenging as we either need to setup component tests with vuex, or handle file downloads. We should do that, but I thnk this change is small and important enough to go in before that. Let me know what you think.

Fixes #189

### DIFF
--- a/pages/download.vue
+++ b/pages/download.vue
@@ -157,7 +157,7 @@
 
                     name: this.datasetName,
                     type: "dataset",
-                    subjects: this.dataTable.annotated.map(row => this.transformAnnotatedRowToSubjectJSON(row))
+                    hasSamples: this.dataTable.annotated.map(row => this.transformAnnotatedRowToSubjectJSON(row))
                 };
             },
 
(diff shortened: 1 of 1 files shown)


jarmoza:
@surchs Reviewing this now...

jarmoza (approved):
Looks good, @surchs. Approved.

Issue #194, opened 2022-09-26
Cancellation of second data table file selection crashes the landing page


jarmoza:
**Steps to reproduce**

1. Select a `participants.tsv` file on the landing page
2. Click the 'Choose File' button once again as if to select another `participants.tsv` file, but instead click 'Cancel' in the file selection dialog box
3. Observe page crash

**Desired behavior**

File selection and file selection cancellation behavior should

1. Not crash the page
2. Be consistent across selection of the participants tsv and json files. For example: File selection displays the file in the corresponding textarea element and file selection cancellation keeps the previously selected file (if any) in the respective store fields and in the corresponding textarea element.

PR #197, opened 2022-10-06
File select cancellation fix + condition cleanup

jarmoza:
This bug fix addresses issue #194. Before, selecting a data table file (`participants.tsv`) and then attempting to select a second file but instead cancelling in the file select dialog resulted in a JS error and page crash.

Once this crash was initially addressed, it also revealed that the code would wipe the first data table selection (as well as its filename and file preview in the UI). 

Neither of these issues occurred for the data dictionary file selection.

Expected behavior:

- [ ] Page does not crash when second data table file selection attempt is cancelled
- [ ] Currently selected data table is not wiped but remains after file select cancellation
- [ ] Data dictionary file selection maintains consistent behavior with data table file selection

The above was addressed by creating a cleaner check against file select event's file data and by not resetting the file data variable if no new file has been selected. (See `onFileSelected` in `file-selector.vue`.)

I also cleaned up some conditional checks for possible values of the file data variable `fileInput` in `file-selector`.  It's either `null` or has a valid value. And I removed the file-select event propagating back up to `index.vue` that was using the string `"none"` to indicate a file selection cancellation.

### DIFF
--- a/components/file-selector.vue
+++ b/components/file-selector.vue
@@ -50,6 +50,7 @@
 
             fileName() {
 
+                // null - No file has been selected yet
                 return ( null === this.fileInput ) ? "" : this.fileInput.name;
             }
         },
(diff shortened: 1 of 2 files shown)


surchs (approved):
Thanks for the PR @jarmoza. This seems to be a MacOS specific bug, as I couldn't reproduce on Linux or Windows. Very interesting.

The changes look good and make sense to me. I'd merge this!

PR #219, opened 2022-11-01
Explanation component test added

### LINKED ISSUE #211: specify annotation-explanation component
## Purpose

Show me an explanation text for what I am supposed to do during the annotation workflow of a specific category, so that I can successfully complete the activity.

## Use cases and input / output

- For the category I am currently annotating, display an explanation text to me in the user interface
  - input:
    - explanation text for the current category (plain string?)
  - output:
- I can click a UI element to collapse / hide the explanation so that I have more screen real estate
  - input and also output:
    - the open / closed state of the explanation component for the current category

## Internal state

- we could also have the open/closed state be stored in the component, but then it would reset every time we navigate away and back to the component.

## Ideas

- the open / closed state could also be something we store globally for all explanation components?
- the explanation component could have the ability to scroll if it exceeds a certain size
- we could also have this be a sidebar element in order to save more vertical screen real estate
- the side-bar could slide in from the side or even be a modal

surchs:
- added component test to illustrate prop behavior minor refactoring:
- use uuid package to ensure random ID instead of passing unique ID as prop
- minor refactor of variable names

closes #211

### DIFF
--- a/components/annot-explanation.vue
+++ b/components/annot-explanation.vue
@@ -5,12 +5,12 @@
         <b-card no-body class="mb-1">
 
             <b-card-header class="p-1" header-tag="header" role="tab">
-                <b-button block v-b-toggle="'accordion-explanation-' + index" variant="info">
-                    {{ uiText.cardTitle }}
+                <b-button block v-b-toggle="'explanation-' + uniqueID" variant="info">
+                    {{ cardTitle }}
                 </b-button>
             </b-card-header>
 
-            <b-collapse :id="'accordion-explanation-' + index" accordion="my-accordion" role="tabpanel">
+            <b-collapse :id="'explanation-' + uniqueID" accordion="explanation-accordion" role="tabpanel">
                 <b-card-body>
                     <b-card-text>{{ explanation }}</b-card-text>
                 </b-card-body>
@@ -23,6 +23,7 @@
 </template>
 
 <script>
+    import { v4 as uuidv4 } from 'uuid';
 
     export default {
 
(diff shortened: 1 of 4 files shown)


components/annot-explanation.vue
```diff
@@ -47,11 +42,11 @@
 
             return {
 
-                // Text for UI elements
-                uiText: {
+                cardTitle: "Explanation",
```
jarmoza, line 45: This is cleaner, but inconsistent with how we've organized UI text in other components - the strings of which are currently contained in a `uiText` object.

Unless we are planning on moving these strings to the store, I suggest going back to `uiText.cardTitle`.

cypress/component/annotation-explanation.cy.js
```diff
@@ -0,0 +1,32 @@
+import annotExplanation from "~/components/annot-explanation";
+
+describe('explanation', () => {
```
jarmoza, line 3: Just a small change to maintain consistent quotation characters throughout. We've gone with double quotation marks elsewhere.

jarmoza (changes requested):
Just two small change requests. Otherwise, feel free to rebase afterwards!

components/annot-explanation.vue
surchs: you're right. Makes sense to me

cypress/component/annotation-explanation.cy.js
surchs: Thanks, good catch. Not sure why this wasn't flagged by the precommit linter.

surchs:
@jarmoza, I worked in both of your comments. If you're good, I'd merge this one with a squash commit because the style stuff doesn't need to be in the history.

jarmoza (approved):
Looks good, @surchs! Go for the squash merge.

PR #222, opened 2022-11-01
Refactor and test missing value component

### LINKED ISSUE #212: specify annotation-missing-values component
## Purpose
For the columns mapped to the category I am currently annotating, show me the unique values that have been declared as missing values, so that I can determine whether they should instead not be missing values.

## Use cases and input / output

-  unique values in columns can be declared as representing "missing" values. for the columns mapped to the category I am currently annotating, show me all values that have been declared "missing". I want to see the value, the column name it belongs to and, if available, a description.
  - input
    - an object with the missing values among columns mapped to the current category under annotation
      - keys are columns
      - values is an array of values for the column that has been declared "missing"
    - a (filtered) data dictionary containing the "levels" attribute for discrete values for the actively mapped columns
      - keys are column names
      - values is the "levels" dictionary where:
        - keys are the value names
        - values are their descriptions (if available)
    - the currently active category (that I am annotating)
- if I want to declare a value to not be missing, I can click on a UI element next to the value (e.g. a button) to remove this value from the missing values
  - input
    - as above
  - output
    - the name of the value to be declared not missing
    - the column that this value belongs to

## Internal State
Component doesn't need internal state

surchs:
Missing value component:

- receives currently active category (e.g. "age") via prop (from a parent component)
- receives a filtered view of missing values from the store via getter
- can look up the description of a value in the store via getter (as before)
- informs the store of a change to the missing values via a mutation call to the store

Component test:
- checks that the component likes the data provided to it
- ensures that component actually calls the mutation with the right parameters

closes #212

### DIFF
--- a/components/annot-missing-values.vue
+++ b/components/annot-missing-values.vue
@@ -15,7 +15,8 @@
                     <template #cell(not_missing)="data">
                         <b-button
                             variant="danger"
-                            @click="removeColumn(data.item)">
+                            :data-cy="'not-missing-button-' + data.item.column + '-' + data.item.value"
+                            @click="removeValue(data.item)">
                             {{ uiText.notMissingButton }}
                         </b-button>
                     </template>
@@ -32,7 +33,7 @@
 <script>
 
     // Allows for reference to store data by creating simple, implicit getters
-    import {mapGetters} from "vuex";
+    import {mapGetters, mapMutations} from "vuex";
 
     export default {
 
@@ -91,11 +92,14 @@
         },
 
         methods: {
+            ...mapMutations([
+                "declareNotMissing"
+            ]),
 
-            removeColumn(p_tableItem) {
+            removeValue(tableItem) {
 
                 // Remove this value from the column's missing value list in the store
-                this.$emit('remove:missingValue', p_tableItem);
+                this.declareNotMissing({column: tableItem.column, value: tableItem.value});
             }
         }
     };
--- a/components/annot-missing-values.vue
+++ b/components/annot-missing-values.vue
@@ -32,15 +32,16 @@
 <script>
 
     // Allows for reference to store data by creating simple, implicit getters
-    import { mapGetters, mapState } from "vuex";
+    import {mapGetters} from "vuex";
 
     export default {
 
         name: "AnnotMissingValues",
 
         props: {
-
-            relevantColumns: { type: Array, required: true }
+            // We need to know what category we are being called for,
+            // so that we can go and ask the store for the correct data
+            activeCategory: { type: String, required: true }
         },
 
         data() {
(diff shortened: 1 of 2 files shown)


components/annot-missing-values.vue
```diff
@@ -32,7 +33,7 @@
 <script>
 
     // Allows for reference to store data by creating simple, implicit getters
-    import {mapGetters} from "vuex";
+    import {mapGetters, mapMutations} from "vuex";
```
jarmoza, line 36: Nice!

cypress/component/annot-missing-values.cy.js
```diff
@@ -28,13 +28,30 @@ describe("missing values", () => {
...
 
-    it.only('handles lack of description gracefully', () => {
-            cy.mount(annotMissingValues, {
-                    propsData: props,
-                    computed: Object.assign(getters, {valueDescription: () => (col, mis) => null })
-                }
-            );
-        }
-    );
+        it('handles lack of description gracefully', () => {
```
jarmoza, line 31: Do we want to standardize the field(s) we're using for mock store fields and functions? I'm referring specifically to how getters is defined at the top of this file as a separate field.

components/annot-missing-values.vue
```diff
@@ -91,11 +92,14 @@
...
+            ...mapMutations([
+                "declareNotMissing"
+            ]),
 
-            removeColumn(p_tableItem) {
+            removeValue(tableItem) {
 
                 // Remove this value from the column's missing value list in the store
-                this.$emit('remove:missingValue', p_tableItem);
+                this.declareNotMissing({column: tableItem.column, value: tableItem.value});
```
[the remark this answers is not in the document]
surchs: You mean instead of the `this.$emit` or `this.$store`? I think clarity and consistency is best. When we use the mapHelpers anywhere, then I think it's better if we use them everywhere - saves one implicit equivalence someone needs to see / understand to not be confused. 

wdyt?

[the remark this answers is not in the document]
jarmoza: This makes sense to me. I didn't remember the `mapMutations` ability too. Happy to stick to one paradigm.

cypress/component/annot-missing-values.cy.js
jarmoza: Yeah. Let's just pick something that mimics what we're injecting into the mount.

`getters`, `actions` (are we doing this?), `mutations`, `props`

surchs: Agreed. And I think `state` is a good idea for passing just the store-state object. Should we put that in the wiki?

Did you want to change something about the tests here before merge too?

surchs: Agreed. And I think `state` is good Should we put that in the wiki?
Also: anything you'd like to change here in the test to match that?

surchs: Ah, I see. If we do wrap it in a big `store` object, we'll need to reference them when passing to the `mount()` directive though, yes? What do you think is the benefit of either you think?

jarmoza: Just the mimicry of the actual app internals. For instance, `props` would not be in the `store` object. But that's about it.

surchs: ah. OK. that makes sense to me!

PR #226, opened 2022-11-11
Split cypress test actions

surchs:
Splits cypress tests into separate runs for 
- component
- end to end

So we can see and check their results separately


jarmoza:
Just going to go ahead and rebase this as the tests work and there are no conflicts to speak of.

PR #229, opened 2022-11-15
Refactor of category-select-table component

jarmoza:
Newly refactored category select table on the categorization page...

1. has required input (`categories`, `categoryClasses`) directly from the store 
2. former `instructions` and `title` props have also been removed, relegated to the new refactored categorization page that will act more like a layout.
3. `selectedCategory` is still emitted to the categorization page (which maintains a `currentCategory` field to share between the `category-select-table` and the `column-linking-table`)

The component test...

1. Mocks store required input via getters
2. Runs through selection of all the default categories.
3. Spies for the emitted selected category output

### DIFF
--- a/components/category-select-table.vue
+++ b/components/category-select-table.vue
@@ -1,58 +1,59 @@
 <template>
 
-    <b-container fluid>
-
-        <!-- Heading for category select component -->
-        <b-row>
-            <h3>{{ title }}</h3>
-        </b-row>
-
-        <!-- Instructions prompting the user how to link categories and columns -->
-        <b-row>
-            <p class="instructions-text">
-                {{ instructions }}
-            </p>
-        </b-row>
+    <div>
 
         <!-- Category selection table -->
-        <b-row>
-            <b-table
-                outlined
-                selectable
-                head-variant="dark"
-                :items="categoryTable"
-                @row-selected="selectCategory"
-                select-mode="single"
-                selected-variant=""
-                :tbody-tr-class="styleTableRow"
-                thead-class="hidden" />
+        <b-row class="no-padding-row">
+
+            <b-col cols="12" class="no-padding-col">
+                <b-table
+                    outlined
+                    selectable
+                    head-variant="dark"
+                    :items="categoryTable"
+                    @row-selected="selectCategory"
+                    select-mode="single"
+                    selected-variant=""
+                    :tbody-tr-class="styleTableRow"
+                    thead-class="hidden" />
+                <b-icon-plus-circle
+                    style="display: block; margin-left: auto; margin-right: auto;"
+                    @click="addCustomCategory" />
+            </b-col>
         </b-row>
 
-    </b-container>
+    </div>
 
 </template>
 
 <script>
 
+    // Allows for reference to store data by creating simple, implicit getters
+    import { mapGetters } from "vuex";
+
     export default {
-
-        props: {
-
-            categories: { type: Array, required: true },
-            categoryClasses: { type: Object, required: true },
-            instructions: { type: String, required: true },
-            title: { type: String, required: true }
-        },
 
         data() {
 
             return {
 
-                selectedCategory: this.categories[0]
+                selectedCategory: ""
             };
         },
 
+        created() {
+
+            // Set the first given category as the selected one
+            this.selectedCategory = this.categories[0];
+        },
+
         computed: {
+
+            ...mapGetters([
+
+                "categories",
+                "categoryClasses"
+            ]),
 
             categoryTable() {
 
--- a/components/category-select-table.vue
+++ b/components/category-select-table.vue
@@ -16,9 +16,6 @@
                     selected-variant=""
                     :tbody-tr-class="styleTableRow"
                     thead-class="hidden" />
-                <b-icon-plus-circle
-                    style="display: block; margin-left: auto; margin-right: auto;"
-                    @click="addCustomCategory" />
             </b-col>
         </b-row>
 
--- a/components/category-select-table.vue
+++ b/components/category-select-table.vue
@@ -30,18 +30,9 @@
 
     export default {
 
-        data() {
+        props: {
 
-            return {
-
-                selectedCategory: ""
-            };
-        },
-
-        created() {
-
-            // Set the first given category as the selected one
-            this.selectedCategory = this.categories[0];
+            selectedCategory: { type: String, required: true }
         },
 
         computed: {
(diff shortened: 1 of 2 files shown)


components/category-select-table.vue
```diff
@@ -1,59 +1,57 @@
...
 
-                selectedCategory: this.categories[0]
+                selectedCategory: ""
             };
         },
 
+        created() {
+
+            // Set the first given category as the selected one
+            this.selectedCategory = this.categories[0];
```
surchs, line 44: We said the categorization page will keep the `selectedCategory` state for both the `category-select` and the `column-linking` component. But inside the `selectedCategory` component we are also keeping a locally scoped `selectedCategory` state variable. Does this mean we are keeping the `"selectedCategory"` state in two places then? 

There are two issues I see with this: 
1. To know what the `selectedCategory` will default to on the parent page, I need to go and look inside the defaults of the component. And then the component has to emit that default choice back to the parent (before any user interaction has happened), so that I can start directly with annotating the column-linking table. 
2. There is a risk of divergence if we're not careful. E.g.: we forget about this double-state and write a process that directly changes the `selectedCategory` state on the parent page (e.g. set it to category 2). Because the parent page doesn't communicate this to the `category-select` component, the component state might now be different (e.g. category 1). And now my UI (highlighted cat1) no longer matches my UX (I'm annotating cat2). 

Long text, short point: I think we should keep state only in one place and have the rest just respond to it! So I would suggest passing `selectedCategory` as a prop to this component and removing the local state.

surchs: Just as a sidenote: I do not actually understand how the component emits its local choice of `selectedCategory` back to the parent without user input. In other words: I do not understand why in the current app version, I can begin annotating with `"Subject ID"` without clicking on the `category-select` table. I just realized that after trying to think about the data flow here.

components/category-select-table.vue
```diff
@@ -1,59 +1,57 @@
...
+        },
+
         computed: {
 
+            ...mapGetters([
+
+                "categories",
+                "categoryClasses"
```
surchs, line 52: Neat!

cypress/component/category-select-table.cy.js
```diff
@@ -0,0 +1,116 @@
...
+        color3: "category-style-2",
+        color4: "category-style-3",
+        color5: "category-style-4",
+        colorDefault: "category-style-default"
+    },
+
+    getters: {
+
+        categories: () => state.categories,
+        categoryClasses: () => state.categoryClasses
```
surchs, line 34: These two getters are the only thing that gets passed to the component. We should put the raw values that are going to be passed into the getters, so that we immediately know what data format the component expects to get from the getters. 

When I write a component test, I shouldn't have to worry about how to store manages to make the getters spit out the data this way, that's not my concern. And secondly, when I look at the test data, it should be easy to see what data will be fed in, I shouldn't have to reason about a factory function or see the test at runtime to get that information.

So concrete suggestion here:
- keep only `getters` in the `store` object, remove the rest
- explicitly set the `getters` to the values that'll get passed at runtime so I can see them. e.g.:

```javascript

    getters: {

        categories: () => [

            "Subject ID",
            "Age",
            "Sex",
            "Diagnosis",
            "Assessment Tool"
        ],
        categoryClasses: () => {
            return {
            "Subject ID": "category-style-0",
            "Age": "category-style-1",
            "Sex": "category-style-2",
            "Diagnosis": "category-style-3",
            "Assessment Tool": "category-style-4"
            };
        }
    }
```
- remove the function `setupCategoryClasses` and its invocation

If you like, I can send a patch or commit to your PR branch with the suggestion.

cypress/component/category-select-table.cy.js
```diff
@@ -0,0 +1,116 @@
...
+            },
+
+            plugins: ["bootstrap-vue"]
+        });
+
+        // Test each row in the category select table
+        for ( let index = 0; index < state.categories.length; index++ ) {
+
+            // 2. Act
+            cy.get(`.category-style-${index}`).click();
```
surchs, line 108: Hehe, I guess we'll have to keep reminding each other about the cypress selectors for a bit. So rather than style, I'd say let's select by string content (i.e. category name) here for a bit more robustness.

surchs (changes requested):
Hey @jarmoza: really cool PR :tada: and a very good size of changes and unit of work! This is great!
I checked it out, confirmed it works. I do have two comments:
1. A minor one about state keeping -> I think whoever holds state for `currentCategory`, it should only be in one place (see detailed comments inline)
2. More important: the tests should be as simple as we can, and only concerned with the component. So I'd say: let's get rid of the store methods and unused state variables and just write explicitly what the getters will give to the component at runtime (again, more detail in comments).

components/category-select-table.vue
jarmoza: @surchs Thanks for the comments!

To address your first point. Thanks for taking the time to think this through, but I think this is one of those cases where if you see something in a PR that's odd/off, push it back to the submitter right away. In this case, I 100% agree with you. In fact, I'm not quite sure why `selectedCategory` is not being passed in as a prop. My guess is that I'm unused to working on components separate from their integration. I am currently on the categorization page refactor itself and yes, the categorization page is where this value will sit (until if/when we ever decide to move it to the store).

>I do not understand why in the current app version, I can begin annotating with "Subject ID" without clicking on the category-select table. I just realized that after trying to think about the data flow here.

The answer to this is in the `data` of the `category-select-table via `selectedCategory: this.categories[0]`. This might in fact be an inconsistent implementation that we have just overlooked where this field was sitting in two places but not causing any issues in behavior.

cypress/component/category-select-table.cy.js
jarmoza: Good comment. No, your meaning is entirely clear. In fact, for the `column-linking-table` test I had evolved my thinking to remove all such "factory" functions for store values. I will just remove that and make the values explicit as suggested.

cypress/component/category-select-table.cy.js
jarmoza: Yeah. I see what you mean. Makes the test that much more explicit. My thinking tends toward genericizing things - in this case for some point if we ever change up categories. Seems reasonable to explicitly select by element text.

components/category-select-table.vue
jarmoza: Resolved with https://github.com/neurobagel/annotation_tool/pull/229/commits/56cab8c016e5c0c30f4d82fd84585210918f7985

cypress/component/category-select-table.cy.js
jarmoza: Resolved with https://github.com/neurobagel/annotation_tool/pull/229/commits/edea6979d53578997f8adecdded201ff65fed02c

jarmoza:
@surchs I believe these last two commits, we are good to rebase this PR with https://github.com/neurobagel/annotation_tool/tree/dev_components_talk_to_store_directly

PR #230, opened 2022-11-17
Column linking table refactor + component test

jarmoza:
**Refactor changes**

1. All props but selected category eliminated. Uses `mapState` instead
2. Table column labels and column to category table are no longer props
 - The labels are defined in the component's data 
 - Table column labels aligned with other components using `uiText` object
 - The creation of the column to category table is done when the component is created (see `setupColumnToCategoryTable` - which has been moved from the categorization page)

**Component tests**

Tests two scenarios to make sure the correct data is being emitted by click on the table rows:
1. Linking a category to a column
2. Unlinking a column from a category


components/column-linking-table.vue
```diff
@@ -18,23 +19,138 @@
...
+
+                "linkColumnWithCategory",
+                "unlinkColumnFromCategory"
+            ]),
+
             applyCategory(p_row, p_index, p_event) {
 
-                // Tell the parent page that a column has been linked with a category
-                this.$emit("column-name-selected", { column: p_row.column });
+                const dataForStore = { column: p_row.column };
```
surchs, line 80: If I understand correctly here: when I click on a column, this `applyCategory` gets called. And it then tells the store: "hey, someone clicked on this column while this category was active - can you go figure this out?". So I think we can leave it up to the store to figure out whether the meaning of this click is going to be a "unlink" or a "link" action. That is, I think here we should just call the store mutation/action

components/column-linking-table.vue
```diff
@@ -18,23 +19,138 @@
...
+                "categoryClasses",
+                "columnToCategoryMap",
+                "dataDictionary",
+                "dataTable"
+            ])
         },
 
         methods: {
 
+            ...mapActions([
```
surchs, line 72: any reason these need to be actions instead of mutations? are you planning on using actions for any multi-mutation event? Or do you expect there to be some asynchronous logic to happen here?

components/column-linking-table.vue
```diff
@@ -18,23 +19,138 @@
...
+                if ( this.selectedCategory !== this.columnToCategoryMap[p_row.column] ) {
+
+                    this.linkColumnWithCategory({...dataForStore, category: this.selectedCategory});
+                } else {
+
+                    this.unlinkColumnFromCategory(dataForStore);
+                }
+
+                // 2. Tell the categorization page a column has been clicked
+                this.$emit("column-name-clicked", dataForStore);
```
surchs, line 92: Why do we need this emit here? At this point I have already told the store about the column click. Could whoever is listening to this emit not receive that information from the store directly?

components/column-linking-table.vue
```diff
@@ -18,23 +19,138 @@
...
+            },
+
+            setupColumnToCategoryTable() {
```
surchs, line 95: I get what this method does. But given how many assumptions we're making here about how the store will be organized, I think it might be cleaner to just ask the store to provide the data elements we want here directly - like we're doing in other components. E.g. one getter to get the column names of the data table, and a second getter to get the description for each column if there is one. 

Then we don't have to do so much work in the component and also have an easier time if we decide to change the specifics in the store.

cypress/component/column-linking-table.cy.js
```diff
@@ -0,0 +1,228 @@
...
+            color3: "category-style-2",
+            color4: "category-style-3",
+            color5: "category-style-4",
+            colorDefault: "category-style-default"
+        }
+    },
+
+    actions: {
+
+        dispatch: (p_actionName, p_payload) => { }
```
surchs, line 100: I really like this. I haven't added the actions to my tests but I think it helps to document them explicitly. One thing though: shouldn't there be two actions here, `linkColumnWithCategory` and `unlinkColumnFromCategory`?

cypress/component/column-linking-table.cy.js
```diff
@@ -0,0 +1,228 @@
...
+    },
+
+    getters: {}
+};
+
+store.getters = {
```
surchs, line 106: why are the getters from above overwritten here?

cypress/component/column-linking-table.cy.js
```diff
@@ -0,0 +1,228 @@
...
+        const onColumnNameClickedSpy = cy.spy().as("onColumnNameClickedSpy");
+        cy.spy(store.actions, "dispatch").as("dispatchSpy");
+        cy.mount(ColumnLinkingTable, {
+
+            mocks: {
+
+                $store: store.actions
+            },
+
+            computed: store.getters,
```
surchs, line 136: I don't think the getters are used by the component at all. Is this an oversight here?

surchs (changes requested):
Hey @jarmoza, thanks a lot for the PR :tada: . Again, really good size of change, very nice to review!

I can confirm that the changes run and pass locally (I guess we already know this now that the CI is working). I do have three main comments (details in inline replies): 
1. I think there is logic in the component that could be refactored into a state getter. For example, you're assuming that a number of store.state variables exist, access them directly, and then pick the data you need from across them. That's a lot of work, I think a getter could do the heavy lifting here (e.g. get you the columns, and a second to get you the descriptions). We have other components who need similar things, and so we can start reusing getters.
2. In some places the data flow isn't completely clear to me. There are actions called in response to click-events, but the component figures out what action to call given the current store state (I think that'd be better handled by the store itself), and then a second emit is done directly to the parent. Maybe that'd be something we can briefly chat about in case I don't get it.
3. Like in https://github.com/neurobagel/annotation_tool/pull/229, I think we should really have less information about specifics in the store / store data in the tests. So ideally we could just hard-code the data that the getters are supposed to provide to the component during testing, and then not have to model the entire store state that is currently mostly used to fill the getters.

cypress/component/column-linking-table.cy.js
jarmoza: The mock store and getters have been completely reworked as per our discussion in Slack huddle.

See https://github.com/neurobagel/annotation_tool/pull/230/commits/01377f20aa87029aef545dad0ddc73b7c9baf76b

components/column-linking-table.vue
jarmoza: `applyCategory` now calls the action (to be implemented in store refactor) `alterColumnCategoryRelation`. This will take a payload of the column and category and determine whether the column should be linked with this category (if it has no category, or a different category associated with it) or if it should be unlinked (the input category is the same as the one it is currently associated with).

I am leaving in the `emit` for now as I would like to handle the general move of next page access to the layout/store in a separate PR.

components/column-linking-table.vue
jarmoza: As discussed in our huddle, the current function of this emit is to let the parent page know to check if the next page (annotation page) is ready to be accessed given the state of the store.

I am leaving in the `emit` for now as I would like to handle the general move of next page access to the layout/store in a separate PR.

components/column-linking-table.vue
jarmoza: As per our discussion of refactoring/writing tests toward an ideal store, this function is removed. Its logic will be split apart/reworked and added to the store. (This work will be added as a subtask of the new store refactor issue I will be creating.)

Instead, now the table rows are computed property that returns an array of objects with keys for column name, column description, and category. The first two pull from a to-be-implemented `columns` store getter that returns an array of objects containing a column name and description per array entry.

The unused column of `category` has also been added to the table (and is now utilized) as another visual indicator of the link between column and category. See `data().uiText.tableFields`.

cypress/component/column-linking-table.cy.js
jarmoza: `linkColumnWithCategory` and `unlinkColumnFromCategory` will be converted to mutations and a new action `alterColumnCategoryRelation` has been added that will call (in future store refactor) either depending on the given action input and state of the store field `columnToCategoryMap`.

jarmoza:
@surchs Take a look, but I think https://github.com/neurobagel/annotation_tool/commit/01377f20aa87029aef545dad0ddc73b7c9baf76b addresses all of the changes we discussed yesterday.

components/column-linking-table.vue
```diff
@@ -18,23 +19,75 @@
...
+
+                    category: this.selectedCategory,
+                    column: p_row.column
+                };
+
+                // 1. Link or unlink the currently-selected category and the clicked column
+                this.alterColumnCategoryRelation(payload);
+
+                // 2. Tell the categorization page a column has been clicked
+                this.$emit("column-name-clicked", payload);
```
surchs, line 90: Did you still want to remove this emit here, now that the action will handle things?

surchs (approved):
Thanks @jarmoza, I think this looks great. One question: did you mean to keep the `emit` in there? I may have misunderstood but I thought the record of the categorization action would be recorded in the store because of the dispatched action. Could you take a look if you still need that `emit` (and the test for it)?

Otherwise, good to go.

components/column-linking-table.vue
jarmoza: @surchs I was thinking of leaving the emit in place and removing it when I do the change over for next page accessibility with the layout/store (See my earlier [comment](https://github.com/neurobagel/annotation_tool/pull/230#discussion_r1029811986).

I suppose it doesn’t matter when the removal happens, but I thought removing it in this PR would be mixing issues.

jarmoza: I was torn over it, and see the logic of both sides. Let's break things. I'll at least leave a `NOTE` in its place with a final commit and then merge.

PR #242, opened 2022-12-05
Refactor annotation-tab component 

### LINKED ISSUE #207: specify annotation-tab component
Purpose: Generic container for primary components of each annotation tab (explanation, columns for annotation, missing values, specialized annotation component)

Actions / User flows:
N/A

Required input:
* activeCategory: the tab needs to know what category it is handling, so it can pass that via prop to its own children
* annnotationComponent: I need to know whether I am handling annotation for a categorical or continuous variable. I should be able to ask the store what annotation component I am supposed to load for the user.

Generated output:
* none

Internal state:
* none

surchs:
Took over this PR from @jarmoza because this component was the last small component to be refactored (rest are pages and I/O things)

Note: the component tests are visually empty because the `annot-tab` component is just a container component for the components displayed on the annotation page. 

The component now has no internal logic or state because it's child components talk directly to the store. The only thing it is passing along is the prop for the `activeCategory`.

Closes #207

### DIFF
--- a/components/annot-tab.vue
+++ b/components/annot-tab.vue
@@ -1,21 +1,21 @@
 <template>
 
-    <div>
+    <div :data-cy="activeCategory">
 
         <!-- Explanation text for this annotation tab -->
-        <annot-explanation :data-cy="'annot-expl-' + title" />
+        <annot-explanation :data-cy="('annot-expl-' + activeCategory)" />
 
         <!-- Lists all the columns linked to the category of this annotation tab -->
-        <annot-columns
-            :active-category="`age`" />
+        <annot-columns :data-cy="('annot-col-' + activeCategory)" :active-category="activeCategory" />
 
         <!-- Lists any values determined to be missing (e.g. potentially invalid) in this tab's columns -->
-        <annot-missing-values />
+        <annot-missing-values :data-cy="('annot-missing-' + activeCategory)" :active-category="activeCategory" />
 
         <!-- Component specializing in the particular kind of annotation for this tab's category -->
         <component
-            :is="annot-explanation"
-            :data-cy="'annot-component-' + title"
+            :is="getAnnotationComponent(activeCategory)"
+            :data-cy="(getAnnotationComponent(activeCategory) + '-' + activeCategory)"
+            :active-category="activeCategory"
             :id-field="idField"
             :title="title" />
 
@@ -30,25 +30,22 @@
 
     export default {
 
+        props: {
+            activeCategory: { type: String, required: true }
+        },
+
         data() {
 
             return {
 
-                details: "",
-                title: "hello",
-
-                // Category of the current annotation tab
-                category: "",
-
-                // String name of the column assigned the 'Subject ID' category in the data table
-                idField: "",
-
-                // Disabled state of the save annotation button
-                saveButtonDisabled: true
+                title: "hello"
             };
         },
 
         computed: {
+            ...mapGetters([
+                "getAnnotationComponent"
+            ])
         }
 
     };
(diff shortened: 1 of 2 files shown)


components/annot-tab.vue
```diff
@@ -1,36 +1,23 @@
 <template>
 
-    <div>
+    <div :data-cy="activeCategory">
 
         <!-- Explanation text for this annotation tab -->
-        <annot-explanation :explanation="details.explanation" :index="details.id" :data-cy="'annot-expl-' + title" />
+        <annot-explanation :data-cy="('annot-expl-' + activeCategory)" />
```
jarmoza, line 6: @surchs Given how the other two child components receive `activeCategory` as props, the `annot-explanation` should receive it as well for when we each annotation component has a different set of instructions.

components/annot-tab.vue
```diff
@@ -44,129 +31,15 @@
...
         },
 
         computed: {
-
             ...mapGetters([
-
-                "getColumnsOfCategory"
-            ]),
-
-            filteredDataTable() {
```
jarmoza, line 80: @surchs Can you briefly account for where these computed property values are now stored in the refactor? In the store? Specialized components? Etc. (`filteredDataTable`, `relevantColumns`, `uniqueValues`, and `uniqueValuesToSubjectMap`).

components/annot-tab.vue
```diff
@@ -1,36 +1,23 @@
...
-    <div>
+    <div :data-cy="activeCategory">
 
         <!-- Explanation text for this annotation tab -->
-        <annot-explanation :explanation="details.explanation" :index="details.id" :data-cy="'annot-expl-' + title" />
+        <annot-explanation :data-cy="('annot-expl-' + activeCategory)" />
 
         <!-- Lists all the columns linked to the category of this annotation tab -->
-        <annot-columns
-            :active-category="details.category"
```
jarmoza, line 10: @surchs Currently, information needed for specific annotation tabs and their components are stored in `annotationDetails` in the store or pulled from the former computed properties of `annot-tab`. Presumably in the refactor, each child component knows how to get the information it needs from the store now based solely on the `activeCategory` passed to `annot-tab`. 

Sound correct?

jarmoza (changes requested):
@surchs One small change and a few other comments just for discussion of the refactor of this component.

components/annot-tab.vue
surchs: Aha, yes, great catch. That's a mistake and I'll fix it!

components/annot-tab.vue
surchs: Yes, exactly. The components are now getting everything that is global state (most things with a few exceptions) directly from the store, either via getters or by directly reading state objects.

components/annot-tab.vue
surchs: I think the general answer is that the child components are getting their needed data directly from the store now (via getters or direct state access), so there is no more need for these prop chains. 

I think the `filteredDataTable` is now built from the `getUniqueValues` getter draft (not implemented yet) in the `annot-categorical` and from the `getPreviewValues` getter draft (not implemented) in the `annot-continuous-values` component. `relevantColumns` is now coming direct from the store via the `getMappedColumns` draft for the `annot-columns` method. `uniqueValues` is also coming direct from the store. 

Might have missed a few. But the overall idea is: nothing except the `activeCategory` prop has to be passed down to the child components, because with just that information they will go to the store and get themselves their appropriate information

jarmoza (approved):
Looks good, @surchs. Go ahead and squash/merge.

PR #258, opened 2022-12-14
Removal of nextPageAccessible getter

jarmoza:
Upon review, it's apparent that the `nextPageAccessible` getter can be removed from the store and its usage in the `next-page` component.

This same functionality can be accessed by calling getter `pageAccessible` with the return value of getter `nextPage` as an argument.


surchs (approved):
Hi @jarmoza: cool, looks good to me.

PR #284, opened 2022-12-16
Implement getCategories store getter

### LINKED ISSUE #254: create getter to get existing category names
- [x] when called, returns an array with the string names of all  categories (built-in and user generated) that currently exist in the store

Other requirements:

surchs:
Closes #254

### DIFF
--- a/components/category-select-table.vue
+++ b/components/category-select-table.vue
@@ -2,7 +2,6 @@
 
     <div>
 
-        <!-- Category selection table -->
         <b-row class="no-padding-row">
 
             <b-col cols="12" class="no-padding-col">
@@ -30,6 +29,8 @@
 
     export default {
 
+        name: "CategorySelectTable",
+
         props: {
 
             selectedCategory: { type: String, required: true }
@@ -39,14 +40,14 @@
 
             ...mapGetters([
 
-                "categories",
+                "getCategories",
                 "categoryClasses"
             ]),
 
             categoryTable() {
 
                 // Return a list of dicts for each category in the table
-                return this.categories.map((name) => ({ category: name }));
+                return this.getCategories.map((name) => ({ category: name }));
             }
         },
 
--- a/components/category-select-table.vue
+++ b/components/category-select-table.vue
@@ -40,14 +40,14 @@
 
             ...mapGetters([
 
-                "getCategories",
+                "getCategoryNames",
                 "categoryClasses"
             ]),
 
             categoryTable() {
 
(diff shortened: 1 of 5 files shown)


components/category-select-table.vue
```diff
@@ -39,14 +40,14 @@
 
             ...mapGetters([
 
-                "categories",
+                "getCategories",
```
jarmoza, line 43: Just for the sake of simplicity - why include `get` in the name? Is this something we just went with from the Miro? The code using the returned value from the getter should not care that it's interacting with a getter.

Since `categories` is now a more complex field in the refactor, and we really want the names of the categories here how about naming this getter something like `categoryNames`?

store/index-refactor.js
```diff
@@ -2,11 +2,16 @@
 import Vue from "vue";
 
 export const state = () => ({
-    columnToCategoryMapping: {}
+    columnToCategoryMapping: {},
```
jarmoza, line 5: We should alphabetize object keys if we don't have a functional grouping for them.

jarmoza (changes requested):
Just two small issues regarding naming/styling and good to go

rmanaem (approved):
Looks good to me! 🎉

components/category-select-table.vue
rmanaem: I think we can take a page out of the old Java OOP best practices and use get(for getters) or set(for mutators) prefixes. It has a nice ring to it but I agree on the 2nd point maybe rename it to getCategoryNames?

surchs: I think I like the combo here. @jarmoza: I like that the methodname tells you in easy language what it does. I haven't found any official recommendations so far, but I like the thinking in this issue: https://github.com/vuestorefront/vue-storefront/issues/2069

jarmoza: @surchs Seems like a good basis for a standard.

store/index-refactor.js
surchs: Good catch @jarmoza. Didn't we have something setup to complain / fix this issue as part of the linting? Should we?

components/category-select-table.vue
surchs: Good point. I'm tracking this here now: https://github.com/neurobagel/annotation_tool/issues/287. For the refactor I would say we make the vuex method names consistent as we implement them

surchs:
nice...

jarmoza (approved):
Looks good @surchs!

Issue #288, opened 2022-12-21
How can we make sure every component test gets a fresh mocked vuex store


surchs:
To mock the vuex store in component tests, at the moment we define the mocked store object before any test is run, and then pass the same reference to the `cy.mount` command of each test in sequence, e.g. 
https://github.com/neurobagel/annotation_tool/blob/86bf385d98aad17b042a17d9d2efe6a5c33d48dd/cypress/component/column-linking-table.cy.js#L93-L105

I just encountered a problem where running two tests that pass individually (e.g. via `it.only`) in sequence fails the second test. The specific problem for me was that I was trying to wrap a function on the mocked `store` object in a `cy.spy` wrapper like so
https://github.com/neurobagel/annotation_tool/blob/86bf385d98aad17b042a17d9d2efe6a5c33d48dd/cypress/component/column-linking-table.cy.js#L92
but if the test was run after a previous test, this failed with `Attempted to wrap undefined property commit as function`

I was able to resolve this specific problem by wrapping the entire mocked store object in a `beforeEach` statement so that each test should get the same `store` object and possible changes don't persist.

But I am not sure that this is the correct way to ensure that there is no interdependence between our tests. The [vuex docs on testing](https://vuex.vuejs.org/guide/testing.html) are a bit sparse, but they seem to use a factoryMethod to create a new store for every test. I am not sure what that would look like for our tests, but I think we should take a look at it and then decide whether we should also use a factoryMethod. See also our [vue-test-utils guide](https://v1.test-utils.vuejs.org/guides/using-with-vuex.html).

jarmoza:
So something like this?

```
function generateStore() { return { ... }; }

cy.mount(ColumnLinkingTable, { 
  
     mocks: { 
  
         $store: generateStore() 
     }, 
     ....
});
```

surchs:
Interesting. Yeah, maybe like that! I think one of the nice things about a store factory method would also be that it is a quick way to get a fresh store with the default initialized state values whenever you need it. 

I am still not completely clear how you would continue mocking things inside of a component test then. Call the factory method to make a fresh store object. And then manually write to state objects (close to what we already have) or run some fancy script to call mutations to make the store look like what you want (close to what we had in the past)? 

I think it would be very nice to see or ask  how others are doing this and then decide whether we also want to copy that here.

jarmoza:
[...]

PR #289, opened 2022-12-21
Implement getColumnDescription getter

### LINKED ISSUE #253: create getter to get column description
This getter has one argument: `column` (required). When called, the getter will return the description for `column`, if it exists and otherwise will return a default value (like empty string)

Other requirements:

surchs:
There is a bit more in this PR than just the normal refactor. The reason is that adding a new test for the 
`column-linking-component` somehow broke the existing component test in a way I didn't understand.
The original `column-linking-component` test passed when forced to run alone `it.only`, but could not wrap the
`commit` function on the `store` object in a `cy.spy` when it was run second after the new test. 

I think this may be because running the first test somehow changes the mocked `store` object in a way
that affects subsequent test. I was not able to debug this though, the `store` object looked normal to me.
As a workaround, I have moved the mocked store into a `beforeEach` statement to make sure every test
gets passed the same `store` object. 

There may be better ways of doing this, and I think we should research these separately. I made https://github.com/neurobagel/annotation_tool/issues/288 to track this.

Closes #253

### DIFF
--- a/components/column-linking-table.vue
+++ b/components/column-linking-table.vue
@@ -53,6 +53,7 @@
 
                 "categoryClasses",
                 "columns",
+                "getColumnDescription",
                 "columnToCategoryMap"
             ]),
 
@@ -62,7 +63,7 @@
 
                     category: this.columnToCategoryMap[column],
                     column: column.name,
-                    description: column.description
+                    description: this.getColumnDescription(column.name)
                 }));
             }
         },
--- a/cypress/component/column-linking-table.cy.js
+++ b/cypress/component/column-linking-table.cy.js
@@ -4,91 +4,104 @@ import ColumnLinkingTable from "~/components/column-linking-table.vue";
 // Documentation for testing Vue events of components
 // https://docs.cypress.io/guides/component-testing/events-vue
 
-// Mocks
-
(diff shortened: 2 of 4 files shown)


jarmoza:
@surchs Other than discussion of the best way to provide unique instances of the mock store in #288, this PR looks good to me. Check out my suggestion in #288 otherwise I think this is okay to go – the `beforeEach` solution is reasonable and with precedent for testing.

jarmoza (commented):
@surchs Other than discussion of the best way to provide unique instances of the mock store in https://github.com/neurobagel/annotation_tool/issues/288, this PR looks good to me. Check out my suggestion in https://github.com/neurobagel/annotation_tool/issues/288 otherwise I think this is okay to go – the beforeEach solution is reasonable and with precedent for testing.

surchs:
Thanks @jarmoza. I like your comments on #288, but I would wait until we have a decision on this before I implement it. Would you mind still approving the PR - I can technically merge without, but I think we should probably not make a habit of it.

jarmoza (approved):
Got it, @surchs. Approved.

rmanaem (approved):
Good to go!

PR #290, opened 2022-12-22
Implement value description getter

### LINKED ISSUE #281: create getter to get description for column values
This getter has two argument: `column` and `value`. When called with both arguments, the getter will return the description for a `value` in `column`, if this description exists and otherwise will return a default value (like empty string).

Other requirements:

surchs:
Closes #281

### DIFF
--- a/store/index-refactor.js
+++ b/store/index-refactor.js
@@ -3,7 +3,8 @@ import Vue from "vue";
 
 export const state = () => ({
     categories: {},
-    columnToCategoryMapping: {}
+    columnToCategoryMapping: {},
+    dataDictionary: {}
 
 });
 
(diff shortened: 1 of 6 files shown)


store/index-refactor.js
```diff
@@ -3,16 +3,25 @@ import Vue from "vue";
...
 });
 
 export const getters = {
 
     getCategoryNames (p_state) {
         return Object.keys(p_state.categories);
+    },
+    getValueDescription (p_state, p_columnName, p_value) {
+        // Returns the description of a value in a column, if that description exists
+        // Otherwise it returns an empty string
```
rmanaem, line 18: Nicely done!

rmanaem (approved):
Good to go!

Issue #259, opened 2022-12-14
create mutation to create column to category map


jarmoza:
when we load a new data table, this mutation should create a new mapping of all the column names to categories, and initialize these mappings with a `null` or default value. 

Other requirements:

- [ ] depending components have been checked and are calling this mutation correctly
- [ ] have unit tests to check functionality
- [ ] needed store state variables exist already or have been created if not

Issue #255, opened 2022-12-14
create getter to get existing column names


surchs:
- [x] when called, returns an array of strings of all column names in the provided data table (or in the data dictionary if equivalent)

Other requirements:

- [x] depending components have been checked and are calling this getter correctly
- [x] have unit tests to check functionality
- [x] needed store state variables exist already or have been created if not

rmanaem:
@surchs Did you work on this? 
The PR that's still open refers to #254 which makes sense but the branch name has 255 in it, maybe a typo?

surchs:
Ooh, my bad @rmanaem. I did start working on this and then I thought I should probably deal with some other stuff first and so I ended up unassigning myself and didn't open a PR. What I should have also done is uncheck the boxes here. I think after all that confusion I also used a misleading branch name for my open PR, so that's what happened... :crab: 

Feel free to take this one! I'll pay a bit more  attention to my issues in the future.

jarmoza:
@rmanaem Might remove this from [my PR for creating the column to category map](https://github.com/neurobagel/annotation_tool/pull/291) (and turn it into a mock), but I've put an implementation of this there.

rmanaem:
@jarmoza Thanks for letting me know. I changed my implementation of the getter to what you had in your PR.

PR #291, opened 2022-12-22
createColumnToCategoryMap mutation + required getter + test

jarmoza:
This PR addresses #259, the creation of a mutation that creates the column to category map. 

@surchs There are a few new complexities with the introduction of this mutation for our process of adding each new element for the store refactor.

Going through the subtasks in #259:

> depending components have been checked and are calling this mutation correctly

The trouble here is that the index page requires `createColumnToCategoryMap`. This requires a check with the index page's old e2e test which is currently broken due to page access (nav item for categorization does not start out as disabled).

> have unit tests to check functionality

Unit tests have been created for the `createColumnToCategoryMap` mutation and also the required, new underlying getter `getColumnNames`

> needed store state variables exist already or have been created if not

`dataTable` has been added to `index-refactor` as it is required by the `columnToCategoryMapping`. (This name should be adjusted when we get a chance minus the -ping).

The addition of the getter `getColumnNames` is the first time we have required a getter in addition to a state field. The issue is what gets passed to the new mutation/getter in question (in this case, `createColumnToCategoryMap`). Since there is some function mapping magic that occurs in the actual Vuex store, calling `p_state.<getterName>` is possible. However, in using our current store mock it is not.


cypress/unit/store-mutation-createColumnToCategoryMap.cy.js
```diff
@@ -0,0 +1,38 @@
...
+                {
+                    "column1Name": "value3", "column2Name": "value4"
+                }
+            ],
+
+            getColumnNames: p_state => ["column1Name", "column2Name"]
+        };
+    });
+
+    it("Returns a list of columns from the data table", () => {
```
surchs, line 27: So far we have made different tests for each getter and mutation. So this test should be moved to its own test file. I think it could make sense to merge all mutations or all getters. But we should discuss and change that in a different PR if we are going to do that.

store/index-refactor.js
```diff
@@ -58,5 +70,12 @@ export const mutations = {
             p_state.columnToCategoryMapping[columnName] = targetCategory;
         }
 
+    },
+
+    createColumnToCategoryMap(p_state) {
+
+        // Column to category map lists all columns as keys with default value of null
+        p_state.columnToCategoryMapping =
+            Object.fromEntries(p_state.getColumnNames().map((column) => [column, null]));
```
surchs, line 79: `getColumnNames` is a getter inside `store.getters...`, you can't reach it in the `store.state`. Mutations get the `store.state` object as their first argument, but not the whole `store`. That is by design although I don't fully get the reasoning (see https://www.github.com/vuejs/vuex/issues/684). I think there are two options here:

1. Have the mutation expect an argument with the list of columnNames. So far the only place we'll call this from is the (future) `action` that is triggered by uploading a new dataTable. So it would make sense for this action to provide the list of category names in the payload. And that would also remove the need to dependence on `dataTable` already being accessible in the `state` - which is a bit odd.
2. Turn this thing into an action instead. Actions "receive a context object which exposes the same set of methods/properties on the store instance", so you can do the whole `context.getters...` thing (see https://vuex.vuejs.org/guide/actions.html).

Between the two, I am much in favour of 1. because it seems a lot cleaner to me. But let me know what you think.

store/index-refactor.js
```diff
@@ -58,5 +70,12 @@ export const mutations = {
             p_state.columnToCategoryMapping[columnName] = targetCategory;
         }
 
+    },
+
+    createColumnToCategoryMap(p_state) {
```
surchs, line 75: based on #259 this guy is supposed to set all mappings to `null`, probably only when the user loads a new `dataTable`. I think `resetColumnToCategoryMap` or something like that (`initialize`?) would probably be more explicit about the intention of this thing. The current name makes me think that the functionality could be very similar to `alterColumnCategoryMapping` which it is not.

cypress/unit/store-mutation-createColumnToCategoryMap.cy.js
```diff
@@ -0,0 +1,38 @@
...
+
+                {
+                    "column1Name": "value1", "column2Name": "value2"
+                },
+                {
+                    "column1Name": "value3", "column2Name": "value4"
+                }
+            ],
+
+            getColumnNames: p_state => ["column1Name", "column2Name"]
```
surchs, line 23: > There are a few new complexities with the introduction of this mutation

Uff, ok, I tink I see what you mean. First: good catch, this is some tricky stuff. There are two things about this: 
1. Mutations get the `state`, not the `store`. So you can't access a `getter` from a mutation. The only reason this works here is because you are mocking it as a `state` method which isn't allowed. I commented more on the store side about how we might be able to address this.
2. So far we only made some simple `getters` that relied on the `state` and nothing else. This is probably OK for testing `getters` and `mutations` (because they only get the state anyway), but if we go towards `actions`, we have to be careful that we don't end up reimplementing half the store in every mock just to make things work. I'm mostly mentioning this here in case we decide to reimplement the mutation you created as an action (which I don't think we should do). In that case, we should probably make another issue and see if we can access the store directly in these tests (without having to mock all of the mutations and getters again).

So, long story short: this getter should go.

cypress/unit/store-mutation-createColumnToCategoryMap.cy.js
```diff
@@ -0,0 +1,38 @@
+import { getters, mutations } from "~/store/index-refactor";
+
+var store = {};
+
+describe("createColumnToCategoryMap", () => {
+
+    beforeEach(() => {
+
+        store = {
```
surchs, line 9: You should rename this (and the initial declaration above) to `state`, because that's what you are mocking here. Alternatively you could be explicit and mock `store.state`. We haven't done that in the tests so far because all they ever needede access to was the `state`. But I'm wondering if we should go back and change that, because it is a bit confusing.

cypress/unit/store-mutation-createColumnToCategoryMap.cy.js
```diff
@@ -0,0 +1,38 @@
...
+    it("Returns a list of columns from the data table", () => {
+
+        const results = getters.getColumnNames(store);
+        expect(results).to.eql(["column1Name", "column2Name"]);
+    });
+
+    it("Creates a column to category map based off the data table", () => {
+
+        mutations.createColumnToCategoryMap(store);
+        expect(store.columnToCategoryMapping).to.eql({"column1Name": null, "column2Name": null});
```
surchs, line 36: neat

cypress/unit/store-mutation-createColumnToCategoryMap.cy.js
```diff
@@ -0,0 +1,38 @@
+import { getters, mutations } from "~/store/index-refactor";
+
+var store = {};
```
surchs, line 3:
```suggestion
let store = {};
```
to be consistent with the rest of the codebase

store/index-refactor.js
```diff
@@ -27,6 +33,13 @@ export const getters = {
             return "";
         }
     },
+
+    getColumnNames(p_state) {
+
+        // Returns list of columns from the loaded data table
+        return ( 0 === p_state.dataTable.length ) ? [] : Object.keys(p_state.dataTable[0]);
```
surchs, line 40: Is there any valid path for this getter to be called before a `dataTable` has been loaded?

surchs (changes requested):
Thanks for the PR @jarmoza. Pretty cool how much stuff can be in 60 changed lines - this one is pretty tricky.

The main thing is that you can't access `getters` from inside `mutations`. I wasn't aware of this either before I tried to understand why your test looked the way it did (yay for testing :tada:). But because of the way things are, we need to make a decision on how we want to implement this. Take a look at my inline comments, and then we should probably get on a call with @rmanaem to make a decision.

As a sidenote: you have addressed both #259 and #255. We should discuss how we handle these situations where we discover an overlapping dependency after we have already assigned the issues so we don't duplicate the work.

store/index-refactor.js
jarmoza: This is not necessarily a straightforward question to answer at the moment because our refactor of the pages for the annotation tool is still in progress. The most direct answer is that it depends on how a getter such as this is used and where – and how gracefully we want to handle failures if the components we hand invalid values to do not.

If a user loads something like the categorization page directly – from a previous browser session for instance and Vue server is still up and running but with a wiped Vuex store – a check like this offers a more graceful failure than what would occur if an empty `dataTable`'s 0th index is accessed.

With the conditional check, the empty array would be offered to the `column-linking-table` component's Bootstrap table. Otherwise the table would be handed `undefined` (or the page itself would crash).

cypress/unit/store-mutation-createColumnToCategoryMap.cy.js
jarmoza: We're not there yet re: merging. As I mentioned in #292 recently, I will be removing this in an incoming commit.

store/index-refactor.js
surchs: Ah ok, that's a really good point. So there is at least one way we know how this getter could be called before a dataTable has been loaded (the one you described). In that case my follow-up question is: shouldn't we then also test in the components that they are in fact OK with receiving these empty arrays? Not for this PR. But if we say: you can reload the page and we'll make sure that the page doesn't crash and instead give you a valid but empty UI, then we should test that this is in fact true. 

But for now, agreed that the check makes sense here.

jarmoza: >shouldn't we then also test in the components that they are in fact OK with receiving these empty arrays? Not for this PR. But if we say: you can reload the page and we'll make sure that the page doesn't crash and instead give you a valid but empty UI, then we should test that this is in fact true.

Absolutely true. Relevant component tests and should test for invalid input if such a scenario exists. And I would say this would extend to page tests, so we can give our users some helpful feedback as to why something has gone wrong and how they might fix it.

cypress/unit/store-mutation-createColumnToCategoryMap.cy.js
jarmoza: >We haven't done that in the tests so far because all they ever needed access to was the state. But I'm wondering if we should go back and change that, because it is a bit confusing.

Let's keep it for now, but I have created an issue for this [here](https://github.com/neurobagel/annotation_tool/issues/295).

store/index-refactor.js
jarmoza: We went with option 1 after meeting between myself, @surchs, and @rmanaem.

store/index-refactor.js
jarmoza: I'm going to go with 'initialize' as 'reset' doesn't quite make sense on its first (and predominant) usage.

jarmoza:
@surchs I think this is ready to go with my latest commit. However, I have changes for `index.vue` that accompany this. I believe these changes will be necessary for your `saveDataTable` and `saveDataDictionary` action refactor tasks. Changes include making sure the data table and data dictionary saving methods in the component have unique names as well as the example where `initializeColumnToCategoryMap` gets called in those methods with data table columns passed in.

For your task that latter call will be need to be moved to the `saveDataTable` action.

surchs (approved):
Looks good @jarmoza. I left one more comment on `deep.equal` vs `equal`. Take a look, but I think we can also discuss that some other time. Either way, good to go!

cypress/unit/store-mutation-initializeColumnToCategoryMap.cy.js
```diff
@@ -0,0 +1,47 @@
...
+        };
+    });
+
+    it("Creates a column to category map based off the data table", () => {
+
+        // Act
+        mutations.initializeColumnToCategoryMap(state, ["column1Name", "column2Name"]);
+
+        // Assert
+        expect(state.columnToCategoryMapping).to.eql({"column1Name": null, "column2Name": null});
```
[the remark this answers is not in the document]
jarmoza: @surchs In this case there is a bit of implicitness going on. Since I'm using `eql` instead of `equal` the comparison will be a deep one.

https://www.chaijs.com/api/bdd/#method_eql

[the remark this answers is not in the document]
surchs: Oh wow, thanks for pointing this out. I had no idea.

PR #292, opened 2022-12-23
Implemented getColumnNames getter 

### LINKED ISSUE #255: create getter to get existing column names

rmanaem:
Closes #255

### DIFF
--- a/cypress/component/column-linking-table.cy.js
+++ b/cypress/component/column-linking-table.cy.js
@@ -37,13 +37,12 @@ describe("The column-linking-table component", () => {
                     };
                 },
 
-                columns: () => {
+                getColumnNames: () => {
 
                     return [
-
-                        { name: "participant_id" },
-                        { name: "age" },
-                        { name: "sex" }
+                        "participant_id",
+                        "age",
+                        "sex"
                     ];
                 },
 
@@ -98,7 +97,8 @@ describe("The column-linking-table component", () => {
     it("can alter link relation (add/remove) between a column and a category", () => {
 
         // 0. The first category and column
-        const participantIDColumn = store.getters.columns()[0].name;
+        const participantIDColumn = store.getters.getColumnNames()[0];
+        console.log(participantIDColumn);
         const subjectIDCategory = store.getters.categories()[0];
 
         // 1. Arrange - Set up the spy, mount the component, and bind the spy to it
--- a/cypress/component/column-linking-table.cy.js
+++ b/cypress/component/column-linking-table.cy.js
@@ -98,7 +98,6 @@ describe("The column-linking-table component", () => {
 
         // 0. The first category and column
         const participantIDColumn = store.getters.getColumnNames()[0];
-        console.log(participantIDColumn);
         const subjectIDCategory = store.getters.categories()[0];
 
         // 1. Arrange - Set up the spy, mount the component, and bind the spy to it
--- a/cypress/unit/store-getter-getColumnNames.cy.js
+++ b/cypress/unit/store-getter-getColumnNames.cy.js
@@ -1,0 +1,20 @@
+import { getters } from "~/store/index-refactor";
+
+describe("getColumnNames", () => {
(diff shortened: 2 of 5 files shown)


cypress/component/column-linking-table.cy.js
```diff
@@ -98,7 +97,8 @@ describe("The column-linking-table component", () => {
     it("can alter link relation (add/remove) between a column and a category", () => {
 
         // 0. The first category and column
-        const participantIDColumn = store.getters.columns()[0].name;
+        const participantIDColumn = store.getters.getColumnNames()[0];
+        console.log(participantIDColumn);
```
surchs, line 101: leftover debug?

cypress/unit/store-getter-getColumnNames.cy.js
```diff
@@ -0,0 +1,20 @@
+import { getters } from "~/store/index-refactor";
+
+describe("getColumnNames", () => {
+    it("Returns an array of all column names in the data table", () => {
+        const { getColumnNames } = getters;
+    const state = {
```
surchs, line 6: indentation looks odd to me

package.json
```diff
@@ -8,7 +8,7 @@
     "start": "nuxt start",
     "generate": "nuxt generate",
     "lint": "npx eslint --ext .vue  --ext .js .",
-    "prepare": "husky install",
+    "lint-fix": "npx eslint --ext .vue  --ext .js . --fix",
```
surchs, line 11: was this done by accident? the `lint-fix` script looks cool, but that should done in a separate PR. And we probably don't want to remove the `prepare` script unless we know it's not needed.

store/index-refactor.js
```diff
@@ -35,7 +36,12 @@ export const getters = {
             return "";
         }
         return description;
+    },
+
+    getColumnNames(p_state) {
+        return (0 === p_state.dataTable.length) ? [] : Object.keys(p_state.dataTable[0]);
```
surchs, line 42: I mentioned the same thing in #291: is there any reason to expect that this `getter` is called before `dataTable` is set to something non-empty? If yes, we should decide whether that's something that should work or not break. And then we should test that our components are still happy if they get an empty array back. 

My sense is that we don't need to worry about this unless there is a good reason this getter could be called without `dataTable` being set. But let me know if you disagree.

surchs (changes requested):
Thanks for the PR @rmanaem. This all looks very good. However #291 has implemented essentially the same getter (minus the tests and component fixes). So I think we should have a chat with @jarmoza on how to proceed with this. 

My sense is that we can probably merge #291 without implementing this getter, but for the moment I would wait for us to chat :).

cypress/component/column-linking-table.cy.js
rmanaem: Yea good catch

rmanaem:
> Thanks for the PR @rmanaem. This all looks very good. However #291 has implemented essentially the same getter (minus the tests and component fixes). So I think we should have a chat with @jarmoza on how to proceed with this.

Yea, I saw Jonathan's PR and comment on #255 when working on this so I replaced my implementation with his to avoid unnecessary conflicts. If I recall correctly he also wrote a test for the getColumnNames getter.

> My sense is that we can probably merge #291 without implementing this getter, but for the moment I would wait for us to chat :).

I also think we should remove(or move) the getColumnNames test from the createColumnToCategoryMap unit test, but let's discuss both of these issues with Jonathan.

jarmoza:
@rmanaem I agree and will remove the `getColumnNames` test from my PR.

surchs (approved):
:tada: good to go! :eyes:

store/index-refactor.js
rmanaem: Answered by Jonathan in #291

Issue #265, opened 2022-12-14
create mutation to set the current page


jarmoza:
The next page button needs to be able to ask the store about what the next page should be, so when I click it I am routed to the next page. This is the mutation to set the current page in order to enable a getter to answer this question.

Other requirements:

- [x] depending components have been checked and are calling this mutation correctly
- [x] have unit tests to check functionality
- [x] needed store state variables exist already or have been created if not

PR #296, opened 2023-01-05
setCurrentPage mutation + unit test for store refactor

jarmoza:
This is a straightforward implementation for both the mutation and test. The mutation itself is currently only used by the `tool-navbar` and the `next-page` components.

This addresses #265


rmanaem (approved):
Looks great @jarmoza! Good to go 👍🏼

Issue #257, opened 2022-12-14
create action to save the uploaded dataDictionary to the store


surchs:
Now that we have said in #256 that loading a new dataTable signals that the app should be completely wiped, this action should expect that a dataTable has already been loaded and there already is a skeleton dataDictionary in the store. 

So the goal here would be: 
- take the uploaded data dictionary
- see what keys already exist in the state.dataDictionary 
- augment existing keys without (ideally) overwriting any already existing content
- ignore all keys that only exist in the uploaded data dictionary but not in the state.dataDictionary
  - whenever a new dataTable is uploaded (e.g. #256), the app state is wiped and re-initialized
  - as part of this, the `state.dataDictionary.annotated` is also re-initialized
  - it will be a skeleton data dictionary that only contains the keys for each column in the `dataTable`, and some barebones attributes to contain future annotations
  - so when a user adds an already created data dictionary, it should get merged into the skeleton data dictionary that was created during `dataTable` upload
  - because we know that the skeleton data dictionary will contain keys for every column in the `dataTable`, if the user supplied data dictionary has additional keys, then they must be faulty or otherwise mismatched
  - in such a case, we should not add them to the data dictionary and we could tell the user that we ignore these keys

surchs:
@rmanaem want to take a look at this?

surchs:
@jarmoza: I updated the general points up there, let me know what you think.

jarmoza:
@surchs Can you explain this use case a bit more so I can understand why we should be ignoring parts of newly uploaded data?

> ignore all keys that only exist in the uploaded data dictionary but not in the state.dataDictionary

jarmoza:
>augment existing keys without (ideally) overwriting any already existing content

@surchs Can you further explain what this sentence means, particularly the word 'augment', here? I need clarification.

surchs:
> augment existing keys without (ideally) overwriting any already existing content

Hey @jarmoza. I think we can drop this requirement. My original idea was: if the user starts with just the table uploaded, that creates a skeleton data dictionary (so annotation can commence). If the user then does a bit of annotation but then decides to come back and upload an existing data dictionary (because they forgot), this data dictionary should be merged with the skeleton data dictionary so none of the already made annotations are lost. 

I now think this is overcomplicated and not required. So what we should do instead is:

1. if the user uploads a dataTable (and app state gets wiped), a skeleton data dictionary is made so annotation can commence
[...]

PR #312, opened 2023-01-16
Data dictionary refactor for Store refactor + Tests

jarmoza:
Here's the final piece of phase 1 of our refactor of the store.

**Refactor Implementation**

As described in #257, this PR implements two store methods: the action `processDataTable` and mutation `setDataTable`. (`processDataTable` utilizes the getter `getColumns` for the update that occurs to the data dictionary via `setDataTable`.)

It is understood that initialization of the data dictionary will only occur with the upload of a new data table via action `processDataTable` fired via file selection on the index page. This creates a skeleton dictionary that only lists blank entries for each of the columns defined in the uploaded 

Any new column in the uploaded data dictionary will be ignored as this was not in the original data table. All values for extant columns in the original data table will however be overwritten. (This re: @surchs 's [latest comment](https://github.com/neurobagel/annotation_tool/issues/257#issuecomment-1384309931).) That latter ability does leave open the possibility that a user may accidentally overwrite annotations. That is a 'let-the-user-beware' ability, but can and should probably be discussed further here.

**Tests**

1. Ignore new columns in a newly uploaded data dictionary (`setDataDictionary` will only be called via `processDataDictionary` - which will be fired via file selection on the index page.)
2. Ensures simple string to string key-value pairs for columns are updated/added from a newly uploaded data dictionary
3. Ensures all keys are retained and new ones added/updated for string to object key-value pairs (and the objects' underlying string to string key-value pairs).

### DIFF
--- a/cypress/unit/store-mutations-setDataDictionary.cy.js
+++ b/cypress/unit/store-mutations-setDataDictionary.cy.js
@@ -1,0 +1,73 @@
+import { mutations } from "~/store/index-refactor";
+
+let state = {};
+
+describe("setDataDictionary", () => {
+
+    beforeEach(() => {
+
+        // Setup
+        state = {
+
+            dataDictionary: {
+
+                "age": {
+
+                    "Description": "age of the participant",
+                    "Units": "years"
+                },
+                "sex": {
+
+                    "Description": "sex of the participant as reported by the participant",
+                    "Levels": {
+                        "M": "male",
+                        "F": "female"
+                    }
+                }
+            },
+
+            dataTable: [
+
+                {
+                    "age": "value1", "sex": "value2"
+                },
+                {
+                    "age": "value3", "sex": "value4"
+                }
+            ]
+        };
+    });
+
+    it("Are new key/value pairs in the uploaded dictionary not added to the data dictionary?", () => {
+
+        // Act
+        mutations.setDataDictionary(state,
+            { "group": { "key1": "value1" } },
+            ["age", "sex"]);
+
+        // Assert
+        expect(state.dataDictionary).to.not.contain.keys("group");
+    });
+
+    it("Has every new key/value pair at least one level deep in the uploaded dictionary been added to the state dictionary, if the key was already present in the state dictionary", () => {
+
+        // Act
+        mutations.setDataDictionary(state,
+            { "age": { "Units": "minutes" } },
+            ["age", "sex"]);
+
+        // Assert
+        expect(state.dataDictionary.age["Units"]).to.equal("minutes");
+    });
+
+    it("Has every new key/value pair two levels deep in the uploaded dictionary been added to the state dictionary, if the key was already present in the state dictionary", () => {
+
+        // Act
+        mutations.setDataDictionary(state,
+            { "sex": { "Levels": { "NB": "Non-binary" }} },
+            ["age", "sex"]);
+
+        // Assert
+        expect(state.dataDictionary.sex["Levels"]).to.contain.keys("NB");
+    });
+});
--- a/cypress/unit/store-mutations-setDataDictionary.cy.js
+++ b/cypress/unit/store-mutations-setDataDictionary.cy.js
@@ -38,7 +38,7 @@ describe("setDataDictionary", () => {
         };
     });
 
-    it("Are new key/value pairs in the uploaded dictionary not added to the data dictionary?", () => {
+    it("New columns in the uploaded dictionary should not be added to the state data dictionary?", () => {
 
         // Act
         mutations.setDataDictionary(state,
@@ -49,7 +49,7 @@ describe("setDataDictionary", () => {
         expect(state.dataDictionary).to.not.contain.keys("group");
     });
 
-    it("Has every new key/value pair at least one level deep in the uploaded dictionary been added to the state dictionary, if the key was already present in the state dictionary", () => {
+    it("New string/string key/value pairs for columns should replace values in the state data dictionary", () => {
 
         // Act
         mutations.setDataDictionary(state,
@@ -57,17 +57,19 @@ describe("setDataDictionary", () => {
             ["age", "sex"]);
 
         // Assert
-        expect(state.dataDictionary.age["Units"]).to.equal("minutes");
+        expect(state.dataDictionary["age"]["Units"]).to.equal("minutes");
     });
 
-    it("Has every new key/value pair two levels deep in the uploaded dictionary been added to the state dictionary, if the key was already present in the state dictionary", () => {
+    it("New string/object key/value pairs in the uploaded dictionary should update/retain values in the state data dictionary", () => {
 
         // Act
         mutations.setDataDictionary(state,
-            { "sex": { "Levels": { "NB": "Non-binary" }} },
+            { "sex": { "Levels": { "M": "male", "F": "female", "NB": "non-binary" }} },
             ["age", "sex"]);
 
         // Assert
-        expect(state.dataDictionary.sex["Levels"]).to.contain.keys("NB");
+        expect(state.dataDictionary["sex"]["Levels"]).to.contain.keys("M");
+        expect(state.dataDictionary["sex"]["Levels"]).to.contain.keys("F");
+        expect(state.dataDictionary["sex"]["Levels"]).to.contain.keys("NB");
     });
 });
--- a/store/index-refactor.js
+++ b/store/index-refactor.js
@@ -35,12 +35,12 @@ export const getters = {
             return "";
         }
     },
-    
+
     getColumnNames(p_state) {
-    
+
         return ( 0 === p_state.dataTable.length) ? [] : Object.keys(p_state.dataTable[0] );
     },
-    
+
     getNextPage(p_state) {
 
         let nextPage = "";
@@ -122,6 +122,11 @@ export const getters = {
 
 export const actions = {
 
+    processDataDictionary( { state, commit, getters }, { data, filename }) {
+
+        commit("setDataDictionary", data, getters.getColumnNames);
+    },
+
     processDataTable( { state, commit, getters }, { data, filename }) {
 
         // This action is dispatched when a new dataTable is loaded by the user.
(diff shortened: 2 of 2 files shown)


cypress/unit/store-mutations-setDataDictionary.cy.js
```diff
@@ -0,0 +1,75 @@
...
+let state = {};
+
+describe("setDataDictionary", () => {
+
+    beforeEach(() => {
+
+        // Setup
+        state = {
+
+            dataDictionary: {
```
surchs, line 12: it should be `dataDictionary.annotated` and `dataDictionary.userProvided`

cypress/unit/store-mutations-setDataDictionary.cy.js
```diff
@@ -0,0 +1,75 @@
...
+
+    it("New string/object key/value pairs in the uploaded dictionary should update/retain values in the state data dictionary", () => {
+
+        // Act
+        mutations.setDataDictionary(state,
+            { "sex": { "Levels": { "M": "male", "F": "female", "NB": "non-binary" }} },
+            ["age", "sex"]);
+
+        // Assert
+        expect(state.dataDictionary["sex"]["Levels"]).to.contain.keys("M");
```
surchs, line 71: why not just assert `deep.equal` of the entire object we expect in `"sex"`?

cypress/unit/store-mutations-setDataDictionary.cy.js
```diff
@@ -0,0 +1,75 @@
...
+
+        // Assert
+        expect(state.dataDictionary["age"]["Units"]).to.equal("minutes");
+    });
+
+    it("New string/object key/value pairs in the uploaded dictionary should update/retain values in the state data dictionary", () => {
+
+        // Act
+        mutations.setDataDictionary(state,
+            { "sex": { "Levels": { "M": "male", "F": "female", "NB": "non-binary" }} },
```
surchs, line 67: Shouldn't we test values that are different from the mocked `dataDictionary` here? 
i.e. `"Levels": { "M": "a male participant" }` or something like that

store/index-refactor.js
```diff
@@ -175,6 +183,21 @@ export const mutations = {
...
     },
 
+    setDataDictionary(p_state, p_newDataDictionary, p_storeColumns) {
+
+        // Update values to existing columns in the data dictionary, but ignore any new columns
+        for ( const column of p_storeColumns ) {
+
+            if ( column in p_newDataDictionary ) {
+
+                for ( const key in p_newDataDictionary[column] ) {
```
surchs, line 193: Is there a reason you're going through the keys in the column one by one here rather than just doing

`p_state.dataDictionary.annotated[column] = Object.assign({}, p_state.dataDictionary.annotated[column], p_newDataDictionary[column])`?

We've been using Object.assing so far for this kind of merging, and it'd be nice to be consistent. So if it is functionally equivalent, we should stick with Object.assign. If `Vue.set` does something `Object.assign` does not do, then we should use `Vue.set` instead.

store/index-refactor.js
```diff
@@ -175,6 +183,21 @@ export const mutations = {
...
+
+                    Vue.set(p_state.dataDictionary[column], key, p_newDataDictionary[column][key]);
```
surchs, line 195: here too: should be `p_state.dataDictionary.annotated`

store/index-refactor.js
```diff
@@ -122,6 +122,11 @@ export const getters = {
 
 export const actions = {
 
+    processDataDictionary( { state, commit, getters }, { data, filename }) {
+
+        commit("setDataDictionary", data, getters.getColumnNames);
```
surchs, line 127: This won't work because the `fileSelector` component returns a string, not an object. See [here](https://github.com/neurobagel/annotation_tool/blob/dev_components_talk_to_store_directly/components/file-selector.vue#L105) and [here](https://github.com/neurobagel/annotation_tool/blob/dev_components_talk_to_store_directly/cypress/component/index_page.cy.js#L126).

I think this in itself might be enough reason to fix up the `fileSelector` component to instead output an object (because that's what you'd expect).

Your call. Either we can parse string -> object in the action and then later refactor the fileSelector component, or we can do it here.

surchs (changes requested):
Thanks for the PR @jarmoza. I left some comments where I believe things need to be changed to align with variable names and returned types. Take a look and let me know what you think.

cypress/unit/store-mutations-setDataDictionary.cy.js
jarmoza: Potato/potata, basically. But if we wanted to change the test input then a `deep.equal` with a locally defined var for the new input makes sense. I'll make the change.

surchs: Yeah, I think it makes more sense if the thing you're passing to the mutation is different in some places from the original dataDictionary. Then we can see whether the merged dataDictionary really updated the different fields and kept the ones that were not specified.

store/index-refactor.js
jarmoza: Let's keep the PR as small as possible as we've mentioned. When fileSelector is refactored and that's the focus we can make the change so that it returns an object.

In the meantime, I'll go with `JSON.parse(data)` here.

store/index-refactor.js
jarmoza: I think maybe you mean, 'Why are you going through key by key?' here? Given the clarified logic it does make sense to assign columns from the `newDataDictionary` but only after confirming that the column exists in the current data dictionary in the store.

cypress/unit/store-mutations-setDataDictionary.cy.js
jarmoza: That would cover a case of 'update' as opposed to just a new key, yes. I'll make the change for `'M'`.

jarmoza:
Okay, @surchs. Take a look at the requested changes I've made. I've also added a little bit more modularity to the tests with the addition of the `columns` and `newDataDictionary` global vars that are based on `state.dataTable` and `state.dataDictionary.provided` that give a similar setup behavior to each test.

store/index-refactor.js
```diff
@@ -168,13 +168,17 @@ export const mutations = {
...
-
         for ( const columnName of Object.keys(p_state.dataTable[0]) ) {
 
             dataDictionary[columnName] = {"description": ""};
         }
 
+        // 2. Create a copy of the data dictionary to save as an original
+        p_state.dataDictionary.provided = Object.assign({}, dataDictionary);
+
+        // 3. Create a copy of the data dcitionary to use for annotation
```
surchs, line 181:
```suggestion
        // 3. Create a copy of the data dictionary to use for annotation
```

store/index-refactor.js
```diff
@@ -168,13 +168,17 @@ export const mutations = {
...
+        // 1. Create a skeleton data dictionary based on the data table's columns
         let dataDictionary = {};
-
         for ( const columnName of Object.keys(p_state.dataTable[0]) ) {
 
             dataDictionary[columnName] = {"description": ""};
         }
 
+        // 2. Create a copy of the data dictionary to save as an original
+        p_state.dataDictionary.provided = Object.assign({}, dataDictionary);
```
surchs, line 179: Good catch about the `provided`, but I think it should go with `setDataDictionary`, no? Currently the dictionary that the user supplies via `processDataDictionary` -> `setDataDictionary` is never stored in `p_state.dataDictionary.provided`. 

I don't disagree with `.provided` also being writter to here, but it definitely must be written to in `setDataDictionary` when the user provided datadictionary is parse.

cypress/unit/store-mutations-setDataDictionary.cy.js
```diff
@@ -11,17 +14,37 @@ describe("setDataDictionary", () => {
 
             dataDictionary: {
 
-                "age": {
+                provided: {
```
surchs, line 17: I don't think anything depends on this part of the datadictionary being mocked? Are you including it only as a way to remind users that this attribute also exists? Or is there something I'm missing?

I think if the goal is to have a pre-defined "new user generated data dictionary" be defined once, then I find it much clearer to have that be done directly in a separate variable like `newDataDictionary`.

cypress/unit/store-mutations-setDataDictionary.cy.js
```diff
@@ -36,40 +59,44 @@ describe("setDataDictionary", () => {
                 }
             ]
         };
+
+        columns = Object.keys(state.dataTable[0]);
+        newDataDictionary = Object.assign({}, state.dataDictionary.provided);
```
surchs, line 64: I find this assignment confusing. Unless I misunderstand, this is the same as `newDataDictionary = state.dataDictionary.provided`, yes? There is no reactivity to worry about here.

As mentioned above, I think it's clearer if you explicitly assign the `newDataDictionary` with the payload you want to reuse across the tests and remove the `.provided` mock.

surchs (changes requested):
hey @jarmoza. Good catch with `.provided`. But I think we need to make sure that the new user-provided data dictionary is actually written to the `.provided` attribute in addition to the `.annotated` one. 

I also left small comments on the tests where I think things could be clearer. LMK what you think.

store/index-refactor.js
jarmoza: Yeah. That's slicker. I see what you mean. Will commit the change in a minute.

cypress/unit/store-mutations-setDataDictionary.cy.js
jarmoza: @surchs What are you referring to by "this part" here?

cypress/unit/store-mutations-setDataDictionary.cy.js
jarmoza: `newDataDictionary = state.dataDictionary.provided` would provide a reference to `state.dataDictionary.provided`. I want a copy of it. In this case, however, it seems we may not be considering that `Object.assign` doesn't deep copy nested objects – I think this is something you realized during the original implementation of the annotation page.

It seems that the way to do this is JSON.parse(JSON.stringify(x));

https://code.tutsplus.com/articles/the-best-way-to-deep-copy-an-object-in-javascript--cms-39655

In any case, I now explicitly define the `newDataDictionary` variable.

cypress/unit/store-mutations-setDataDictionary.cy.js
surchs: @jarmoza i was referring to the 'userProvided' part you had mocked here

cypress/unit/store-mutations-setDataDictionary.cy.js
```diff
@@ -0,0 +1,103 @@
...
+        newDataDictionary["group"] = { "key1": "value1" };
+
+        // Act
+        mutations.setDataDictionary(state, newDataDictionary, columns);
+
+        // Assert
+        expect(state.dataDictionary.annotated).to.not.contain.keys("group");
+    });
+
+    it("New string/string key/value pairs for columns should replace values in the state data dictionary", () => {
```
surchs, line 80:
```suggestion
    it("New string/string key/value pairs for existing columns should replace values in the state data dictionary", () => {
```

surchs (approved):
looks good @jarmoza :tada: 
edit: ah, missed the failing test. let's fix those first

jarmoza:
@surchs A Cypress test is failing on your end?

Here's me:
Screen Shot 2023-01-17 at 10 34 34 AM

surchs:
> 

The tests are failing here in the PR at the moment. I didn't run them locally

jarmoza:
I will revert to my previous implementation of `initializeDataDictionary` since we cannot call a mutation from a mutation. `dataDictionary.provided` and `dataDictionary.annotated` will be initialized directly here.

surchs:
Oh, good catch. It looked like an action because of the collapsed rows.

jarmoza:
@surchs Fixed!

PR #325, opened 2023-01-24
Mutations should have a single payload parameter

### LINKED ISSUE #324: Refactor store mutations to follow deconstruction syntax
Vuex only allows one payload argument per mutation. So if we want to pass multiple arguments, we need to pass an object or deconstruct a passed object directly in the mutation signature.

jarmoza:
Closes #324


surchs (approved):
Looks good @jarmoza !

PR #326, opened 2023-01-24
Swapping in refactored store

### LINKED ISSUE #307: switch over to the refactored store
We want to switch over the new refactored store and replace the current store with it. The old store can be removed.

jarmoza:
Closes #307


surchs (approved):
Looks good @jarmoza

PR #327, opened 2023-01-25
Basic app launch post-refactor

### LINKED ISSUE #308: can we launch a partially complete app
With #251 done, the app should now partially work again - at least up to the categorization page. Here we want to make sure that:

- [ ] we can launch the app even though part of it relies on not-yet-implemented store methods (e.g. annotation page)
- [ ] OR: understand how we can partially launch the app in the state it is in now

This can be closed when the current app state can be successfully launched via `npm run dev` locally, even if the UI or functionality doesn't work correctly yet.

jarmoza:
Closes #308 

The simple addition of field `pageData` back into the store allows the annotation tool to launch.

### DIFF
--- a/store/index.js
+++ b/store/index.js
@@ -5,18 +5,10 @@ export const state = () => ({
 
     categories: {
 
-        "Subject ID": {
-
-        },
-        "Age": {
-
-        },
-        "Sex": {
-
-        },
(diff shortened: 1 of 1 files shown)


surchs (approved):
Nice, looks good @jarmoza.

Issue #311, opened 2023-01-11
adjust categorization page tests


surchs:
make sure the categorization e2e page tests run with the new store. We probably have to remove the assessment-tool related tests because we are removing this functionality until it can get reimplemented (#306).

jarmoza:
Closed by #330

Issue #310, opened 2023-01-11
adjust index page tests


surchs:
Ensure the index page e2e page test runs and passes. If it doesn't pass, check whether the test needs to be rewritten to account for the store refactor.

jarmoza:
Closed by #330

Issue #309, opened 2023-01-11
refactor page test setup


surchs:
Our page e2e tests rely on some cypress support functions to run (e.g. [here](https://github.com/neurobagel/annotation_tool/blob/dev_components_talk_to_store_directly/cypress/e2e/page/categorization-pagetests.cy.js#L15)). These functions need to be refactored to make use of the new store methods (updated actions, mutations). The goal is that the page tests run (not necessarily pass).

jarmoza:
Closed by #330

PR #330, opened 2023-01-30
Adjusting e2e tests + support code for Phase 1 store refactor

jarmoza:
This PR addresses #309, #310, and #311.

**309 - refactor page test setup**

1. Refactored `loadTestDataIntoStore` to account for new store actions for the data table and dictionary as well as the now-required order for file selection (data table -> data dictionary).
2. Added new support function `commitToNuxtStore` to address the need to set the current page when programmatically routing between annotation tool pages. More on that below.
3. Removed ability to only check next page button and not navbar from `assertNextPageAccess` as it is no longer used in any e2e tests. (With the refactor, the annotation page will no longer have UI elements with their own 'next' buttons – where this test functionality was used.)

**310 - adjust index page tests**

1. The second e2e test for the index page now just checks to make sure that the data dictionary file selection is disabled while a data table has not yet been selected.
2. Some light reformatting and test string adjustments.

**311 - adjust categorization page tests**

1. All category tool group tests have been removed from the e2e tests for the categorization page.
2. A multi-category linking test has been added to ensure next page access is maintained after a second non-Subject ID categorization has been done. (Current 'Subject ID' categorization is a special case that grants access to the next page.)
3. Store mutation `setCurrentPage` is now called after routing to the categorization page has been done. Without this direct call, the tool believed the current page to still be the home page after the first call to `cy.visit('/')` and errored in showing what looked to be the enabled next page button for the home page. (The `currentPage` store field is now only set when a navbar item or next page button is clicked on, so mutation `setCurrentPage` needed to be called as part of a programmatic navigation.)
4. Some light reformatting and test string adjustments.


cypress/e2e/page/index-pagetests.cy.js
```diff
@@ -51,26 +51,31 @@ describe("tests on the index page via store interaction", () => {
...
+                    .find("label")
+                    .first()
+                    .should("have.class", "disabled-file-selector-button");
+
+                // 3. Attempt to click on 'Choose file' button for data dictionary
                 cy.get("[data-cy='data-dictionary-selector']")
                     .contains("Choose file")
-                    .click()
-                    .selectFile(dataFolder + p_dataset.data_dictionary);
+                    .click();
```
surchs, line 70: If I understand correctly, this click should not be possible at this stage, right? Is there any way to assert that directly? If not, I'd add a brief explanation why to the comment above - because I was expecting something like `.should("be.disabled")` like we're usually using.

cypress/support/commands.js
```diff
@@ -118,6 +117,16 @@ Cypress.Commands.add("datasetMeetsTestCriteria", (p_pageName, p_datasetConfig, p
     return true;
 });
 
+// Calls mutation in the Nuxt store
```
surchs, line 120: Oh that's neat! Is this directly commiting to the vuex store? Or is the "Nuxt store" a separate thing?

cypress/e2e/page/categorization-pagetests.cy.js
```diff
@@ -36,7 +36,12 @@ describe("tests on categorization page ui via programmatic state loading and sto
                 // when a page is 'visited' by Cypress
                 cy.window().its("$nuxt.$router").then(router => {
 
+                    // A. Route to categorization page
```
surchs, line 39: just out of curiosity: if you had "clicked" on the "Categorization page" nav-bar link, would that come out to the same thing here?

surchs (approved):
Looks good @jarmoza! Very cool moment to see the app actually come back to life after all this refactoring :tada:! All looks good to me, ready to merge. I left two comments more out of curiosity. Depending on your answer for the index-page test one, I would maybe make the comment for the "click data-dictionary button" step a bit clearer regarding ("why cannot we just assert that this label-button thing is not clickable here")

edit:
Oh, forgot an important point: now that the e2e tests can (at least in principle) pass again, should we exclude the annotation page e2e tests until they are reasonably passing? The goal would be to have from now and going forward fully green tests again1

cypress/e2e/page/index-pagetests.cy.js
jarmoza: The `click` itself is possible (producing no response), but the `selectFile` makes the test fail because label that would be clicked on in `file-selector` has been disabled. There is a way to check for a Cypress error (e.g. 'is not clickable'), but apparently [this is highly discouraged](https://github.com/cypress-io/cypress/discussions/21150).

>If not, I'd add a brief explanation why to the comment above - because I was expecting something like .should("be.disabled") like we're usually using.

👍

jarmoza: >Oh, forgot an important point: now that the e2e tests can (at least in principle) pass again, should we exclude the annotation page e2e tests until they are reasonably passing? The goal would be to have from now and going forward fully green tests again!

Yes, the annotation page e2e tests (and really the download page e2e tests too - even though it is sort of passing now) should be fully functional by the end of Phase 2 refactor work.

cypress/e2e/page/categorization-pagetests.cy.js
jarmoza: Yes. This is would be akin to (though not identical to) a more traditional e2e test (i.e. `simple-e2etest.cy.js`) where instead of loading state we navigate through the UI.

I just tried it out and it works!

I am calling support function `nextPageByNav` (which is a wrapper for `cy.get(`[data-cy='menu-item-${p_navItemName}'] a`).click();`) right after the `loadTestDataIntoStore` in the `beforeEach` setup function. This also eliminates the need to make a call to the Nuxt `router` off of the `window` object.

cypress/support/commands.js
jarmoza: Nope. Nuxt store is meant to be synonymous with Vuex store here. I could just change the function names.

But more importantly - yes! We are committing directly to the store here. Before I just had the function to `dispatch` a store action. In this case, the `commit` call became needed since we refactored the store to more often call mutations directly.

cypress/e2e/page/categorization-pagetests.cy.js
surchs: No strong opinion. I think we started this because of the advice on the cypress docs to not set state via the UI. I have to admit I still don't fully understand why that'd be an antipattern for us. Maybe it makes a bigger difference when you do it for external services (e.g. doing a Google-OAuth process via the UI every time you want to login?). 

All else equal, I'd go with the version that you think is more maintainable / easier to reason about when reading the test.

cypress/e2e/page/index-pagetests.cy.js
surchs: Just to be clear about the part with the annotation-test: If we know they won't pass until phase 2 is done, maybe let's disable them for now so the e2e tests can also pass now, and seeing them fail conveys an actual problem as we're finishing phase 2.

cypress/e2e/page/categorization-pagetests.cy.js
jarmoza: I'm going to stick with the programmatic loading of state, but it's good to keep in mind that's an option.

cypress/support/commands.js
jarmoza: Instances of 'NuxtStore' have been changed to 'VuexStore' - including in our testing contribution documentation (https://github.com/neurobagel/annotation_tool/pull/330/commits/f839304315dae34546b3e94c42ec262270d5776c)

Issue #176, opened 2022-08-22
Categorical annotations not in output if saved more than once


surchs:
Steps to reproduce:

- have a "categorical" column (e.g. Sex or Diagnosis)
- annotate the values
- hit "Save"
- make a change to any mapping
- hit "Save again"
- go to Download page

Expected behavior:

- Download page shows the last "saved" mappings

Actual behavior:

- Download page shows empty mappings for the column

**Tests**:
Not sure if these are better tested as component tests for the categorical annotation procedure or e2e of the entire annotation experience.

If I categorize a column as "Sex" (or "Diagnosis") and then completely annotate the values of that column:
- [x] then the store should not be updated (or component shouldn't emit an update event) if I haven't hit "save" yet
- [x] once I hit "save", the store should be updated (and/or component should emit update event with the correct data)
- [x] if I change the annotations after I have already made annotations and hit save a second time (more than once), then the store should reflect the newly updated annotations (or component emit update event with new data)
- [x] when I navigate to the download page after a 1+x annotation save workflow, the download page column should contain the mappings I have just created (i.e. the words should appear, order is irrelevant)

![Image](https://user-images.githubusercontent.com/1302022/185973134-190d03a3-bab1-49eb-99f7-c8e6ba52caf9.gif)

surchs:
I added some test cases that I think would prevent this from being reintroduced down the line. Let me know what you think @jarmoza.

jarmoza:
Thanks @surchs. This issue could be a good basis to now try a component Cypress test. I do believe that writing an "e2e"-ish test is proper as well.

The idea being that once the component test(s) (for likely each of the different annotation tab components) is written, at least one e2e test should be written to ensure this saving of annotation twice shows up properly in the table on the download page.

surchs:
I think that's a good idea for getting more test coverage while you chase this bug. If you can write a e2e / outside / behavioural test that reproduces this bug first, then that would be a great indicator of when a fix is working - and also a good future regression test.

PR #335, opened 2023-02-01
Build static

surchs:
This PR sets up the master branch for deployment on GH pages and to use a custom domain https://annotate.neurobagel.org

Most steps are just direct from the docs: https://nuxtjs.org/deployments/github-pages/

I did some extra things:
- removed the client-only / no-ssr tag on the annotation.vue page (not compatible with static build)
- removed the custom route prefix that is suggested for GH deployments in the docs because we are using a custom domain

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,17 @@ $ npm run generate
 
 For detailed explanation on how things work, check out the [documentation](https://nuxtjs.org).
 
+## Deployment
+
+To deploy the static build on GH pages, run
+
+```bash
+npm run generate
+npm run deploy
+```
+
+See the [Nuxt documention](https://nuxtjs.org/deployments/github-pages/) for more details.
+
 ## Special Directories
 
 You can create the following extra directories, some of which have special behaviors. Only `pages` is required; you can delete them if you don't want to use their functionality.
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -60,5 +60,9 @@ export default {
 		}
 	},
 
-	devtool: "source-map"
+	devtool: "source-map",
+	target: "static",
+	router: {
+		base: "/annotation-tool/"
+	}
 };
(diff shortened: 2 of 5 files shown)


jarmoza (commented):
@surchs Looks good. I just tested out the app. Very cool that this can exist statically.

All of the functionality seems to work with the exception of diagnosis annotation - where there is some wonkiness is some columns not getting annotated or getting wiped after being re-annotated.

Might be worth a brief looking into as I don't recall this problem in the master build.

surchs:
Do you mean like this: https://github.com/neurobagel/annotation_tool/issues/128 ?

jarmoza:
No. Apologies for being unclear where this is happening. I mean on the Download page. I tried annotating for all categories just now (Subject ID, Age, Sex, Diagnosis (Group and GroupDX), Assessment Tools (1 group with IQ). I was seeing one and then both of the Diagnosis columns being completely clear.

Retrying just now I believe I see where the bug is, however. It's when the Diagnosis category has its annotations saved more than once.

This is the result on the download page.

Screen Shot 2023-02-02 at 10 20 53 AM

All that said, I think this is a new-ish bug (similar to #176). I also note that the behavior of the 'Save Annotation' button is inconsistent across tabs (Does it gray out when saved to indicate a save has occurred or does it stay green?). I'll file issues for these two things.

I'm going to go ahead and approve this PR for merge.

jarmoza (approved):

surchs:
Ah yeah,  I think #176 is the better fit, agreed. That said: I think we encountered this behaviour/bug before and discussed that it will be resolved together with the larger refactor. If you think this is a new bug, then we should take another look, but not very urgently as the underlying codebase will (mostly) be replaced once the refactor branch merges.

jarmoza:
@surchs Agreed. We can visit this at a later time. I've filed issues #339 and #340 for them.

Issue #270, opened 2022-12-14
create mutation for setting current heuristic of continuous values column


jarmoza:
This will be written to the data dictionary per column under the key 'transformationHeuristic'.

Other requirements:

- [ ] depending components have been checked and are calling this mutation correctly
- [ ] have unit tests to check functionality
- [ ] needed store state variables exist already or have been created if not

jarmoza:
An open questions is whether or not we should continue to think of categories as being assigned to data types or if such a heuristic should be assigned to a column in a `columns` store field.

jarmoza:
Thanks for catching that, @surchs.

jarmoza:
Closed as of #337

PR #337, opened 2023-02-01
New mutation setHeuristic for phase 2 store refactor

jarmoza:
This PR addresses #270.

1. `setHeuristic` is now a mutation that takes a column and heuristic name and assigns the heuristic to the `transformationHeuristic` field in the data dictionary object for that column
2. A unit test has been created for the new `setHeuristic` mutation.
3. The `annot-continuous-values` component has been amended to understand `setHeuristic` as a mutation instead of an action.

### DIFF
--- a/cypress/unit/store-mutation-setHeuristic.cy.js
+++ b/cypress/unit/store-mutation-setHeuristic.cy.js
@@ -11,13 +11,28 @@ describe("setHeuristic", () => {
 
             dataDictionary: {
                 annotated: {
-                    column1: { transformationHeuristic: "" }
+                    column1: {}
                 }
             }
         };
     });
 
-    it("Set the transformation heuristic of a column", () => {
+    it("Set the heuristic of a column with no transformationHeuristic key", () => {
+
+        // Act
+        mutations.setHeuristic(state, {
+            column: "column1",
+            heuristic: "column1Heuristic"
+        });
+
+        // Assert
+        expect(state.dataDictionary.annotated.column1.transformationHeuristic).to.equal("column1Heuristic");
+    });
+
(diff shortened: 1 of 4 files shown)


cypress/unit/store-mutation-setHeuristic.cy.js
```diff
@@ -0,0 +1,31 @@
...
+describe("setHeuristic", () => {
+
+    beforeEach(() => {
+
+        // Setup
+        state = {
+
+            dataDictionary: {
+                annotated: {
+                    column1: { transformationHeuristic: "" }
```
surchs, line 14: minor point: in the default state of the `store.state` the `transformationHeuristic` attribute doesn't exist for a column. 
That makes sense, not every column will need one, so probably best not to create the field by default. 
However: it would be good to have a test here that checks that we actually successfully create the attribute as well.
i.e.: 

- mock the state to not have `transformationHeuristic` (i.e. to reflect the default state), 
- fire the mutation
- ensure that 1) the respective column now does have the `transformationHeuristic` attribute, and 2) that the value of this attribute is correct

surchs (approved):
Nice, looks good @jarmoza. I left some suggestions for clarity - all minor. Add where you agree, but otherwise good to go!

cypress/unit/store-mutation-setHeuristic.cy.js
jarmoza: @surchs Makes sense. See what you think about my new commit.

jarmoza: @surchs Okay. Just added the existence check as well. Check it out and see what you think.

Issue #336, opened 2023-02-01
Feature: Ability to choose different transform heuristics for columns in continuous values component


jarmoza:
We want to allow users to select different types of data transformation heuristics (from those available) for the columns of a continuous values category (i.e. 'Age').

1. UI changes must be made in the in `annot-continuous-values` component for this to occur. The suggested solution is a multi-tabbed continuous values component where each tab represents a column and where within each tab there is a dropdown for the column's transform heuristic – somewhat similar to the current implementation that selects a heuristic for all columns.
2. Once the choice is made a mutation must be written to alter the annotated data dictionary to register the newly selected heuristic for its respective column.

Tied with #270 and #274

jarmoza:
@surchs Does this sound about right to you?

surchs:
Makes sense to me @jarmoza. What I would add is that both the "user-selected-a-heuristic-for-this-column-please-store" mutation and the "what-heuristic-is-currently-stored-for-this-column" getter have to be updated for this to work. And, as we discussed: the choice of available heuristics (i.e. a list of possible string names of heuristics the user can choose from) is still going to be tied to the category.

surchs:
@jarmoza can we close this?

jarmoza:
@surchs No, this is the new feature that will be my upcoming PR.

jarmoza:
Closed by #372

Issue #266, opened 2022-12-14
create a getter for the active heuristic for a column


jarmoza:
Implement a getter that retrieves the currently applied heuristic for a column (see `transformationHeuristic` in the data dictionary).

Other requirements:

- [ ] depending components have been checked and are calling this getter correctly
- [ ] have unit tests to check functionality
- [ ] needed store state variables exist already or have been created if not

jarmoza:
Closed by #342

PR #342, opened 2023-02-02
New 'getHeuristic' getter for phase 2 store refactor

jarmoza:
This completes #266.

1. getter `getHeuristic` returns the `transformationHeuristic` assigned to a particular column in `dataDictionary.annotated`. If the given column has no `transformationHeuristic` key, a blank string is returned
2. Unit tests in `store-getter-getHeuristic.cy.js` test for both of the above scenarios.
3. The naming of the getter is updated in the `annot-continuous-values` component, `getActiveHeuristic` -> `getHeuristic`. The reasoning here is that 'active' referred to the operative heuristic for a particular category, and now that heuristics can vary across columns this is no longer the case.

Moves us one step closer to #336.

### DIFF
--- a/cypress/unit/store-getter-getHeuristic.cy.js
+++ b/cypress/unit/store-getter-getHeuristic.cy.js
@@ -1,0 +1,45 @@
+import { getters } from "~/store";
+
+describe("The column-linking-table component", () => {
+
+    let store = {};
+
+    beforeEach(() => {
+
+        store = {
+
+            state: {
+
+                dataDictionary: {
+
+                    userProvided: {},
+                    annotated: {
+
+                        column1: {}
+                    }
+                }
+            }
+        };
+    });
+
+    it("Get the heuristic of a column with no heuristic", () => {
+
+        // Act
+        const transformationHeuristic = getters.getHeuristic(store.state)("column1");
+
+        // Assert
+        expect(transformationHeuristic).to.equal("");
+    });
+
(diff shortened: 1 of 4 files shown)


cypress/unit/store-getter-getHeuristic.cy.js
```diff
@@ -0,0 +1,45 @@
+import { getters } from "~/store";
+
+describe("The column-linking-table component", () => {
```
surchs, line 3:
```suggestion
describe("The getHeuristic getter", () => {
```
Did you mean this?

cypress/unit/store-getter-getHeuristic.cy.js
```diff
@@ -0,0 +1,45 @@
...
+        };
+    });
+
+    it("Get the heuristic of a column with no heuristic", () => {
+
+        // Act
+        const transformationHeuristic = getters.getHeuristic(store.state)("column1");
+
+        // Assert
+        expect(transformationHeuristic).to.equal("");
```
surchs, line 31: cool!

surchs (approved):
Looks great @jarmoza! I left one suggestion for the name of the test group, see if that fits.

cypress/unit/store-getter-getHeuristic.cy.js
jarmoza: @surchs  I should say it does! Thanks for catching that.

PR #343, opened 2023-02-03
Only run index & categ e2e tests for now

jarmoza:


surchs (approved):
So nice and green! Very cool @jarmoza :tada:

Issue #275, opened 2022-12-14
create mutation to change 'missing' status of a column's value


jarmoza:
This mutation, `changeMissingStatus`, should add/remove a value from a column's list of missing values in the store.

As per our [Miro refactor draft](https://miro.com/app/board/uXjVP8wIUms=/), a `missingValues` list will be stored in the column objects in `dataDictionary.annotated`.

**NOTE:** Currently, this is referenced in `annot-missing-values` in the more specific form called `declareNotMissing` and in `annot-categorical` as `designateAsMissing`. The latter, however, is itself missing from the current implementation of `annot-continuous-values`.

- [x] depending components have been checked and are calling this mutation correctly
- [x] have unit tests to check functionality
- [x] needed store state variables exist already or have been created if not

jarmoza:
This would be a good candidate for the proposed, new `columns` store field.

jarmoza:
Currently this is referenced in `annot-missing-values` in the more specific forms called `declareNotMissing' and in `annot-categorical` as 'designateAsMissing'. 'designateAsMissing', however, is itself missing from the current implementation of `annot-continuous-values`.

jarmoza:
@surchs Does a 'Mark as missing' button make sense for the continuous values component?

jarmoza:
Yes, I just noticed it was missing but its component test defines an unused `designateAsMissing` mutation (now changed to `changeMissingStatus`). Not planning on implementing it presently.

jarmoza:
Closed by #359

PR #355, opened 2023-02-09
Getter getMissingValues for Phase 2 Store Refactor

### LINKED ISSUE #278: create getter for missing values of a category
This task to create a getter that retrieves the list of missing values for a column.

As per our [Miro refactor draft](https://miro.com/app/board/uXjVP8wIUms=/), a `missingValues` list will be stored in the column objects in `dataDictionary.annotated`.

jarmoza:
This closes #278.

This PR implements a getter for retrieving any values that have been marked as "missing' for a particular given column in the annotated data dictionary. The expectation is that if there are any missing values they will be stored in `dataDictionary.annotated.<columnName>.missingValues`.

Also implemented are unit tests for three scenarios for this getter:
1. Retrieving the missing values list from`dataDictionary.annotated.<columnName>.missingValues`
2. Returning an empty list if there are no missing values (no `missingValues` key) in the column entr in the `dataDictionary.annotated` object
3. Returning an empty list if the column itself is not listed in `dataDictionary.annotated`

This is related to #275 - which will be implemented following this PR.

### DIFF
--- a/store/index.js
+++ b/store/index.js
@@ -158,6 +158,19 @@ export const getters = {
         return mappedColumns;
     },
 
+    getMissingValues: (p_state) => (p_column) => {
+
+        let missingValues = [];
+
+        if ( p_column in p_state.dataDictionary.annotated &&
+             "missingValues" in p_state.dataDictionary.annotated[p_column] ) {
+
+            missingValues = p_state.dataDictionary.annotated[p_column].missingValues;
+        }
+
+        return missingValues;
+    },
+
     getNextPage(p_state) {
 
         let nextPage = "";
--- a/store/index.js
+++ b/store/index.js
@@ -172,14 +172,33 @@ export const getters = {
         return mappedColumns;
     },
 
-    getMissingValues: (p_state) => (p_column) => {
+    getMissingValues: (p_state) => (p_category) => {
 
-        let missingValues = [];
+        // 0. Retrieve all columns linked with the given category
+        const mappedColumns = [];
+        for ( const column in p_state.columnToCategoryMapping ) {
 
-        if ( p_column in p_state.dataDictionary.annotated &&
-             "missingValues" in p_state.dataDictionary.annotated[p_column] ) {
+            if ( p_category === p_state.columnToCategoryMapping[column] ) {
 
-            missingValues = p_state.dataDictionary.annotated[p_column].missingValues;
+                mappedColumns.push(column);
+            }
+        }
+
+        // 1. Build a map of missing values by column
+        let missingValues = {};
+        for ( const column of mappedColumns ) {
+
+            // A. Starts out as a blank list
+            missingValues[column] = [];
+
+            // B. Save a list of missing values for this column if,
+            // 1) the column has an entry in the data dictionary and,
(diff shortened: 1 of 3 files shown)


store/index.js
```diff
@@ -158,6 +158,19 @@ export const getters = {
         return mappedColumns;
     },
 
+    getMissingValues: (p_state) => (p_column) => {
```
surchs, line 161: I think this is the inverse situation to https://github.com/neurobagel/annotation_tool/pull/353: 

We want to ask for "all missing values in any column assigned to a category", so the getter should take a category and then figure out what columns to get the missing values from. 

If you take a look at `annot-missing-values`, the returned value is expected to be an array of tuples (?) with the first entry being the columnName, and the second the missing Value. 

Not saying you need to stick to this exact format, but the general functionality should be as described.

surchs (changes requested):
Looks very nice @jarmoza. As mentioned in the comment, I think this is the inverse situation to the previous PR: as far as I understand, this getter is supposed to take a `"category"` and then go and fetch the missing values from all columns that have been associated with this category. 

The `annot-missing-values` component is (afaik) the only user of this, and it expects a specific format for this - take a look. Make sure to update the getter name in this component as well.

store/index.js
```diff
@@ -172,14 +172,33 @@ export const getters = {
...
-        if ( p_column in p_state.dataDictionary.annotated &&
-             "missingValues" in p_state.dataDictionary.annotated[p_column] ) {
+            // B. Save a list of missing values for this column if,
+            // 1) the column has an entry in the data dictionary and,
+            // 2) if a missing values list for the column has already been made
+            if ( column in p_state.dataDictionary.annotated &&
+                "missingValues" in p_state.dataDictionary.annotated[column] ) {
 
-            missingValues = p_state.dataDictionary.annotated[p_column].missingValues;
+               missingValues[column] = p_state.dataDictionary.annotated[column].missingValues;
```
surchs, line 200: The `annot-missing-values` component expects a different format atm, see: https://github.com/neurobagel/annotation_tool/blob/dev_components_talk_to_store_directly/components/annot-missing-values.vue#L80 (i.e. array of objects for each value, object has column name and value repeated).

Whatever format you pick, make sure that the format the getter produces matches what the component expects.

surchs (approved):
Looks great @jarmoza, thanks for the changes - didn't realize the PR was still open. Good to go :tada: !

store/index.js
jarmoza: @surchs After looking the code over, I believe this implementation is consistent with what the `annot-missing-values` component expects in its `tableItems` function. The only change that needs to occur for `annot-missing-values` is the name change for the getter.

That said, in looking over the `tableItems` function there I'd like to refactor it. The code inside is quite confusing for me to read. I just did so in order to better understand what it's attempting to do - comment description aside. I can add that refactor in a separate PR or here since an adjustment needs to be made to `annot-missing-values` anyway.

jarmoza: I'm just going to make the name change for the getting in `annot-missing-values` and will include a refactor for `tableItems` in a separate PR.

PR #357, opened 2023-02-15
Missing values table items refactor + component test fix

jarmoza:
This PR

1. Simplifies the implementation of the `tableItems` function in the `annot-missing-values` component
1. Fixes the `annot-missing-values` component test that was broken by #355 (a name change for the use of the `getMissingValues` getter in `annot-missing-values`).
2. Formats the `annot-missing-values` component test

### DIFF
--- a/components/annot-missing-values.vue
+++ b/components/annot-missing-values.vue
@@ -74,19 +74,26 @@
             ]),
 
             tableItems() {
-                // Returns an array of objects, with one object for each missing value
-                // in the columns assigned to the activeCategory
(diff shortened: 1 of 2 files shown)


surchs (approved):
Looks good @jarmoza. Let's collect styling related issues that do not impact functionality so we can address them in one PR. Mixing functional and non-functional changes makes the PR and later the git history harder to understand

jarmoza:
@surchs Makes sense to me, re: styling. Will go with that from here on out.

PR #362, opened 2023-02-16
Added and configured lint staged and lint workflow

### LINKED ISSUE #301: Set up a workflow to run the linter on new pull requests
Set up a workflow to run eslint on PRs to catch any issues in case eslint wasn't run before committing.

### LINKED ISSUE #331: Install and configure lint-staged
Same issue as https://github.com/neurobagel/query-tool/issues/30: 

Eslint currently just runs on everything, but we want it to run only on the staged changes. This tool will enable that: https://github.com/okonet/lint-staged

rmanaem:
Closes #301 
Closes #331

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,0 +1,16 @@
+name: lint
+
+on: [pull_request, workflow_dispatch]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+
+    - name: Install dependencies
+      run: npm install
+
+    - name: Run ESLint
+      run: npm run lint
--- a/.lintstagedrc.json
+++ b/.lintstagedrc.json
@@ -1,0 +1,3 @@
+{
+    "*.{js,vue}": ["npm run lint"]
+}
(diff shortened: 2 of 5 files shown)


.github/workflows/lint.yaml
```diff
@@ -0,0 +1,16 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+
+    - name: Install dependencies
+      run: yarn
+
+    - name: Run ESLint
+      run: npm run lint
```
surchs, line 16: could the mix of `npm` for running the `lint` and `yarn` to install the dependencies above become a problem?

.lintstagedrc.json
```diff
@@ -0,0 +1,3 @@
+{
+    "*.{js,vue}": ["npm run lint"]
```
surchs, line 2: can we get it to run on our component tests as well? `*.cy.js`?

surchs (approved):
Thanks for taking this on @rmanaem :tada: . This looks good to go for me. Left a question and a suggestion, see if you agree.

.github/workflows/lint.yaml
rmanaem: I don't think so but I can change `yarn` to `npm install`

.lintstagedrc.json
rmanaem: I believe it does as is

PR #363, opened 2023-02-16
getUniqueValues getter for Phase 2 Store Refactor

### LINKED ISSUE #277: create getter for unique values of a category
This getter, `getUniqueValues` takes a category (and an optional number to limit number of unique values returned) and compiles a list of all unique values in the columns linked to that category – as long as they are *not missing* values.

jarmoza:
This PR closes #277 by implementing the following:

1. `getUniqueValues` getter in the store that takes every column linked to a given category and gathers unique values for each column via the `dataTable`. The getter outputs an object keyed on column with values that are an array of the unique values for that column. Missing values are skipped via lookup in the `dataDictionary`
2. `getUniqueValues` also takes an optional parameter `maxValues` to limit how many values are returned
3. `displayItems` in the `annot-categorical` component has been rewritten to accommodate the new output of `getUniqueValues`
4. Unit tests for this getter in `store-getter-getUniqueValues.cy.js` test the following scenarios: a) unique values for a single column, b) unique values for multiple columns, c) retrieving unique values with the optional maximum amount argument, and d) ensuring that no 'missing' values are returned with the unique values

### DIFF
--- a/store/index.js
+++ b/store/index.js
@@ -243,7 +243,7 @@ export const getters = {
 
     getUniqueValues: (p_state) => (p_category, p_maxValues="None") => {
 
-        // 1. Construct a list of unique values for each column
+        // 1. Construct an object containing a list of unique values for each column
         const uniqueValues = {};
         for ( const columnName in p_state.columnToCategoryMapping ) {
 
@@ -275,7 +275,7 @@ export const getters = {
             }
         }
 
-        // Return a list of objects
+        // Return an object containing a list of unique values for each column
         return uniqueValues;
     },
 
--- a/store/index.js
+++ b/store/index.js
@@ -1,4 +1,5 @@
 // Facilitate Vue reactivity via 'Vue.set' and 'Vue.delete'
+import { Set } from "core-js";
 import Vue from "vue";
 
 export const state = () => ({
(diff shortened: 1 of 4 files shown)


store/index.js
```diff
@@ -240,6 +241,44 @@ export const getters = {
         return p_state.transformationHeuristics[columnDataType];
     },
 
+    getUniqueValues: (p_state) => (p_category, p_maxValues="None") => {
+
+        // 1. Construct a list of unique values for each column
```
rmanaem, line 246: I think this should be `Construct an object containing...` since `uniqueValues` is an object

store/index.js
```diff
@@ -240,6 +241,44 @@ export const getters = {
...
+                // NOTE: Trimming is done here instead of only looking at p_maxValues rows
+                // just in case there are blank entries for columns in the data table
+                if ( "None" !== p_maxValues ) {
+
+                    uniqueValues[columnName] = uniqueValues[columnName].slice(0, p_maxValues);
+                }
+            }
+        }
+
+        // Return a list of objects
```
rmanaem, line 278: I think this should be `Returns an object ...`

rmanaem (approved):
Great work on the PR @jarmoza! The unit-test test is thorough and in-line comments are nice and helpful. I left two comments that simply address the type of `uniqueValues` mentioned in the in-line comments other than that  It's good to go!

store/index.js
jarmoza: Good point. Comment needed to be updated with implementation change.

Issue #364, opened 2023-02-16
Missing values should be added/removed from data dictionary 'missingValues' list


jarmoza:
Currently missing values are not being added/removed from the `missingValues` key under each column object in `dataDictionary.annotated`. Since the UI on the annotation page is not yet fully testable during the phase 2 refactor of the store, this was overlooked.

jarmoza:
Closed by #365

PR #365, opened 2023-02-16
changeMissingStatus getter Fix + Check for Unique Values

### LINKED ISSUE #171: adding value to missing twice should not work, independently of UI
This is an old draft, filing it here as a test case:

> If I mark a column-value as "missing" and then do it again (e.g. click button twice), the second time I try this, something should catch that the value is already declared missing (I actually thought I had a catch for this)

jarmoza:
This PR closes #171  and #364 by implementing the following:

1. Ensuring missing values are added/removed from the `missingValues` list that will exist for each column in `dataDictionary.annotated`
2. Add a uniqueness check to the add value case in `changeMissingStatus`
3. Updating unit tests for `checkMissingStatus` to properly add/remove from `missingValues` lists
4. Additional unit test to ensure no duplicate values can be added to the `missingValues` list in `dataDictionary.annotated.<column>`.

### DIFF
--- a/cypress/unit/store-mutation-changeMissingStatus.cy.js
+++ b/cypress/unit/store-mutation-changeMissingStatus.cy.js
@@ -15,7 +15,10 @@ describe("changeMissingStatus mutation", () => {
 
                     annotated: {
 
-                        column1: []
+                        column1: {
+
+                            missingValues: []
+                        }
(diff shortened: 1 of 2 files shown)


rmanaem (approved):
Good to go!

PR #366, opened 2023-02-17
selectCategoricalOption mutation for Phase 2 Store Refactor

### LINKED ISSUE #274: create a mutation to select an option for the value of a categorical column(s)

jarmoza:
This PR closes #274 by implementing the following:

1. `selectCategoricalOption` mutation in the store that takes a column name, option value and raw value and a) creates a `valueMap` object on `dataDictionary.annotated.<column_name>` if it does not exist, and b) assigns the raw value and option value as a key/value pair in the `valueMap` object.
2. Three unit tests in `store-mutation-selectCategoricalOption.cy.js` that test a) if the value map is created when it does not yet exist, b) if the option (annotated) value is set in the value map for the raw value, and c) ensures a previously entered option value can be overwritten in the value map.
3. `annot-categorical` component and its component test in `annot-categorical.cy.js` are updated to reflect changes for the new `selectCategoricalOption` mutation

### DIFF
--- a/components/annot-categorical.vue
+++ b/components/annot-categorical.vue
@@ -24,7 +24,7 @@
                         <v-select
                             :data-cy="'categoricalSelector' + '_' + row.index"
                             :value="getSelectedOption(row.index)"
-                            @input="selectAnOption($event, row.item['columnName'], row.item['rawValue'])"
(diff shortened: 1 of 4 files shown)


rmanaem (approved):
👨🏼‍🍳 🥯

Issue #272, opened 2022-12-14
create getter for preview of value after heuristic transformation


jarmoza:
Implement a getter, `getHarmonizedPreview`, that returns a version of the original value as transformed by the currently selected transformation heuristic for a category ([soon to be column](https://github.com/neurobagel/annotation_tool/issues/336)).

- [x] depending components have been checked and are calling this mutation correctly
- [x] have unit tests to check functionality
- [x] needed store state variables exist already or have been created if not

jarmoza:
@surchs I think we should have a discussion here in this issue and/or by video regarding the different transform heuristics we have in regard to their implementation pre-refactor – and how we might want to change them.

surchs:
ok, happy to discuss here! Do you want to start by explaining what you'd like to change?

jarmoza:
Closed by #368

PR #368, opened 2023-02-23
getHarmonizedPreview getter for Phase 2 Refactor

jarmoza:
This addresses #272 by implementing the following:

1. The creation of new store getter `getHarmonizedPreview` which takes a column name and a data table value and produces a transformed version of that value given the transformation heuristic assigned to the given column. Of note, for this new implementation of functionality that was contained in the previous component `annot-age-values` is that `isoyear` transformation is currently being left (commented) out so it can be more fully addressed in a more robust way in its own new issue.
2. The introduction of the `appSetting` store field as mentioned in our Miro board draft of the store refactor. Currently this only contains the `missingValueLabel` for the `'string'` transformation heuristic.
3. Tests for each different transformation heuristic have been added in `store-getter-getHarmonizedPreview.cy.js`
4. Correlating adjustments have been made to the `annot-continuous-values` component and its test file `annot-continuous-values.cy.js`

### DIFF
--- a/components/annot-continuous-values.vue
+++ b/components/annot-continuous-values.vue
@@ -53,10 +53,9 @@
 
             ...mapGetters([
 
-                "getPreviewValues",
                 "getHarmonizedPreview",
-                "getTransformOptions",
-                "getHeuristic"
+                "getHeuristic",
+                "getTransformOptions"
             ]),
 
             validationItems() {
--- a/cypress/component/annot-continuous-values.cy.js
+++ b/cypress/component/annot-continuous-values.cy.js
@@ -15,7 +15,7 @@ const store = {
                 "column2": ["2,1", "22,1"]
             };
         },
-        getTransformHeuristics: () => (activeCategory) => {
+        getTransformationHeuristics: () => (activeCategory) => {
             return ["float", "bounded", "euro", "range", "int", "string", "isoyear"];
         }
     },
--- a/cypress/component/annot-continuous-values.cy.js
+++ b/cypress/component/annot-continuous-values.cy.js
@@ -7,12 +7,8 @@ const store = {
 
     getters: {
 
-        getHarmonizedPreview: () => (p_column, p_missingValue) => {
-            return null;
-        },
-        getHeuristic: () => (p_column) => {
-            return null;
-        },
+        getHarmonizedPreview: () => (p_column, p_missingValue) => null,
+        getHeuristic: () => (p_column) => null,
         getPreviewValues: () => (p_activeCategory) => {
 
             return {
--- a/cypress/component/annot-continuous-values.cy.js
+++ b/cypress/component/annot-continuous-values.cy.js
@@ -7,8 +7,12 @@ const store = {
 
     getters: {
 
-        getHarmonizedPreview: () => (p_column, p_missingValue) => null,
-        getHeuristic: () => (p_column) => null,
+        getHarmonizedPreview: () => (p_column, p_missingValue) => {
+            return null;
+        },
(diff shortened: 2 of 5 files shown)


jarmoza:
@rmanaem NOTE: Something is currently preventing our e2e tests from running.

jarmoza:
@rmanaem Just pushed the e2e yaml change in a new commit. Looks like we're in the green now for tests here too.

components/annot-continuous-values.vue
```diff
@@ -53,10 +53,9 @@
 
             ...mapGetters([
 
-                "getPreviewValues",
```
rmanaem, line 56: What's the reason for removing `getPreviewValues`?

cypress/component/annot-continuous-values.cy.js
```diff
@@ -7,18 +7,22 @@ const store = {
 
     getters: {
 
-        getHarmonizedPreview: () => (p_column, p_missingValue) => null,
-        getHeuristic: () => (p_column) => null,
+        getHarmonizedPreview: () => (p_column, p_missingValue) => {
+            return null;
```
rmanaem, line 11: Curious, is this change for improving readability or is the `=> null` syntactically incorrect?

cypress/unit/store-getter-getHarmonizedPreview.cy.js
```diff
@@ -0,0 +1,185 @@
...
+                    }
+                }
+
+            }
+        };
+    });
+
+    it.only("float transformation", () => {
+
+        console.log("Pre-setup");
```
rmanaem, line 36: Are the `console.log` statements left over from debugging?

cypress/unit/store-getter-getHarmonizedPreview.cy.js
```diff
@@ -0,0 +1,185 @@
...
+                            transformationHeuristic: ""
+                        }
+                    }
+                }
+
+            }
+        };
+    });
+
+    it.only("float transformation", () => {
```
rmanaem, line 34: Is the `it.only` left over from developing and debugging or did you want to only run float transformation?

components/annot-continuous-values.vue
jarmoza: So this is an oversight on my part actually. `getPreviewValues` was kind of an imagined getter placed in the initial refactor of this file. At the moment it's necessary because it's called by the `validationItems` function in `annot-continuous-values`, but this will be removed in my next PR for #336 .

cypress/component/annot-continuous-values.cy.js
jarmoza: I was having an issue with the test earlier, and was playing with the syntax to see if Cypress had a problem with this way of defining an arrow function. I believe the answer is that it does not. I can replace the original syntax.

cypress/unit/store-getter-getHarmonizedPreview.cy.js
jarmoza: Thought I removed all of those. Will delete and commit.

cypress/unit/store-getter-getHarmonizedPreview.cy.js
jarmoza: Yes, this is leftover from debugging. Will delete and commit along with the `console.log` removal.

jarmoza:
@rmanaem PR requested changes committed. And tests are good.

rmanaem (approved):
👨🏼‍🍳 🥯

PR #372, opened 2023-03-06
Feature for Phase 2 Refactor: Transform Heuristics for continuous values are now selectable by column

jarmoza:
This PR implements #336.

1. The `annot-continuous-values` component (that is used on the annotation page tab, currently exclusively for 'Age'-categorized columns) now features one tab per column
2. Each tab lists the data for that column with an original value and preview value column
3. Each contains a dropdown featuring all of the app's transformation heuristics for continuous values - including a default blank with which no transformation has yet taken place
4. Unit tests in `annot-continuous-values.cy.js` have been reworked and extended for the new feature

NOTE: @rmanaem There was something happening in the testing that I could not explain. The reason that the `store` variable is initialized each time in the `beforeEach` function is because its function definitions in `store.getters` were seemingly being wiped by (at least) the first individual test. This is difficult to debug because `JSON.stringify` will not display functions defined on an object. Not sure this is necessarily worth our time to looking into, but if you can see something that might have caused that behavior please point it out.

### DIFF
--- a/components/annot-continuous-values.vue
+++ b/components/annot-continuous-values.vue
@@ -6,18 +6,36 @@
         <b-card-header>{{ uiText.instructions }}</b-card-header>
         <b-card-body class="continuous-values-card-body">
 
-            <v-select
-                data-cy="selectTransform"
-                :options="getTransformOptions(this.activeCategory)"
-                :value="getHeuristic(this.activeCategory)"
-                @input="commitHeuristic($event)" />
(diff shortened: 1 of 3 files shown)


rmanaem (approved):
:cook:

jarmoza:
Thanks @rmanaem. Will take a look at that branch's tests and see if I can note any similarities.

Issue #347, opened 2023-02-06
Annotation Page Refactor for Phase 2 Store Refactor


jarmoza:
Now that we are into the [second phase of the store refactor](https://github.com/neurobagel/annotation_tool/issues/305) - moving onto the annotation and download pages - it will be necessary to refactor the annotation page itself before its dependent components can be fully tested manually and via our page-level e2e tests for the page.

Below is a list of the refactor work that needs to be done to the annotation page:

- [x] Tabs now must look at new store configuration for info on their construction (previously in store field `annotationDetails`) #348 
- [x] Column unlinking from categories must be updated - includes a) transforming a column to its original unannotated state and b) the actual unlinking of a column from its previously assigned category #376 
- [x] Removing all old tool group functionality
- [x] Altering how missing values are added and removed to accord with the store refactor #381 
- [x] Altering how the annotations are 'saved' now that changes are being made to `dataDictionary.annotated` instead of the previous `dataTable.annotated`

Once this work is complete, then the e2e test for the annotation can be [re-enabled for the github testing workflow](https://github.com/neurobagel/annotation_tool/issues/341). At this point, the e2e tests for the annotation page will likely need to be updated as well:

- [x] Re-enable annotation page e2e test (See #341.)
- [x] Update annotation page e2e test until all tests pass

jarmoza:
In light of the possible complexity of this page refactor, I will spin off this work into several PRs.

jarmoza:
Blocked by #386

surchs:
This seems to be unblocked / finished. @jarmoza / @rmanaem could you check + close?

rmanaem:
I believe this is finished @jarmoza could you confirm?

jarmoza:
@surchs / @rmanaem Confirmed. Closed.

PR #375, opened 2023-03-13
Phase 2 Store Refactor - Integration Part 1 - Annotation page access, tabs refactor, column-category unlinking

### LINKED ISSUE #348: Annot-tab refactor for Phase 2 Store Refactor
The `annot-tab` component that we use to build the tabs of the annotation needs to be refactor for the new changes to the store introduced for [phase 1](https://github.com/neurobagel/annotation_tool/issues/251) and [phase 2](https://github.com/neurobagel/annotation_tool/issues/305) of our recent store refactor.

It contains several child components - `annot-explanation`, `annot-columns`, `annot-missing-values` and the specialized annotation component for each tab - for which utilization of the Cypress testing stub capability will be necessary.

jarmoza:
In an effort to keep PRs small, this PR closes the first step of restoring the annotation page (access + tabs) now that the second phase of our store refactor is (mostly) complete. As such this PR is not yet testable via our e2e page tests since the annotation page itself is functionally incomplete and its related e2e tests need to also be refactored to adapt to changes we have made. (Closes #348.)

In addition to basic access to the page with its primary components loading, this PR also addresses the following, as noted in #347 : 

- Tabs now must look at new store configuration for info on their construction (previously in store field annotationDetails)

Here is a file by file explanation of the changes made thus far:

`annot-categorical.vue`
- `v-select` value attribute is temporarily disabled as this reflects a getter we missed in our refactor discussions: `getSelectedOption`. @surchs also noticed this and there will be a brief discussion about this in #374 
- Refactors for new parameterization of `changeMissingStatus` and `selectCategoricalOption` getters

`annot-categorical.cy.js`
- Tests updated for refactor and brought in line with how `annot-continuous-values` is tested (re: act/assert calls)

`annot-continuous-values.vue`
- Call to `setHeuristic` mutation brought in line with how `annot-categorical` handles `v-select` input by calling the relevant mutation directly. Removed wrapper function `commitHeuristic`

`annot-continuous-values.cy.js`
- Updated `changeMissingStatus` mock to take object argument its real store version takes

`annot-tab.vue`
- Simple removal of now unused `id-field`

`annotation.vue`
- Annotation tab creation no longer uses removed `annotationDetails` store field but rather `columnToCategoryMapping` (and `colorInfo` for coloring)
- The `annot-tab` child component's props have been greatly reduced here. Keeping commented out version for now for any remaining functionality needed in the next integration PR; specifically for missing values.
- A `categorySkipList` field has been added to the page for generality. Really, our current hardcoded behavior is that no tab should be created for `Subject ID` and this is passed to the new `getMappedCategories` getter
- Keeping commented out store maps and functionality until next integration PR
- See `index.js` below for new criteria created for access to the annotation page

`index.js`
- New getter `getMappedCategories` that looks at `columnToCategoryMapping` and returns a unique list of categories that have thus far been linked to columns in the input data table. A optional list of categories to skip returning is given as a parameter. The existence of this getter is up for discussion, but its creation seemed to make sense as opposed to placing this store functionality inside `annotation.vue`
- Because we have not yet had a discussion on how we want optional data dictionary key value pairs to be created, it is currently necessary to check for the existence of the `missingValues` list in both the `getMappedColumns` getter and `changeMissingStatus` mutation.
- New `isPageAccessible` criterion has been added for moving to the annotation page: `notOnlySubjectIDCategorized`. Discussion point here as well is whether or not we want to allow users to move to the annotation page without categorizing a column other than one as 'Subject ID'. This doesn't make sense if we have no other categorizations as with the current implementation no tabs would exist on the annotation page. In the previous implementation, all categories had hardcoded tabs via `annotationDetails`
- Change `selectCategoricalOption` mutation's parameters to one parameter object to bring it line with the rest of our store mutations
- Fix for missing values check in `getUniqueValues`

`store-getter-getMappedCategories.cy.js`
- New unit tests for new getter `getMappedCategories` for annotation page

`store-getter-isPageAccessible.cy.js`
- Changed test to account for new criteria where at least one column must be linked to a non-'Subject ID' category for the user to proceed to the annotation page

`store-mutation-selectCategoricalOption.cy.js`
- Updates for new object argument for `selectCategoricalOption` mutation

### DIFF
--- a/cypress/e2e/page/categorization-pagetests.cy.js
+++ b/cypress/e2e/page/categorization-pagetests.cy.js
@@ -68,7 +68,7 @@ describe("Tests on categorization page ui via programmatic state loading and sto
                     // 2. Categorize first subject id column in table as 'Subject ID'
                     cy.categorizeColumn("Subject ID", p_dataset["category_columns"]["Subject ID"][0]);
 
-                    // 3. Assert that annotation nav and next page button are enabled
+                    // 3. Assert that annotation nav and next page button are disabled
                     cy.assertNextPageAccess("annotation", false);
                 }
             });
--- a/cypress/e2e/page/categorization-pagetests.cy.js
+++ b/cypress/e2e/page/categorization-pagetests.cy.js
@@ -69,7 +69,7 @@ describe("Tests on categorization page ui via programmatic state loading and sto
                     cy.categorizeColumn("Subject ID", p_dataset["category_columns"]["Subject ID"][0]);
 
                     // 3. Assert that annotation nav and next page button are enabled
-                    cy.assertNextPageAccess("annotation", true);
+                    cy.assertNextPageAccess("annotation", false);
                 }
             });
 
--- a/store/index.js
+++ b/store/index.js
@@ -343,7 +343,8 @@ export const getters = {
 
                     // a. Check to see if this value is marked as 'missing' for this column
                     let value = p_state.dataTable[index][columnName];
-                    if ( !p_state.dataDictionary.annotated[columnName].missingValues.includes(value) ) {
+                    if ( "missingValues" in p_state.dataDictionary.annotated[columnName] &&
+                         !p_state.dataDictionary?.annotated[columnName]?.missingValues.includes(value) ) {
 
                         uniqueValues[columnName].add(value);
                     }
--- a/store/index.js
+++ b/store/index.js
@@ -241,6 +241,12 @@ export const getters = {
             p_state.dataDictionary.annotated[p_columnName].transformationHeuristic : "";
     },
 
+    getMappedCategories: (p_state) => (p_skipCategories=[]) => {
+
+        // Return list of all categories linked to columns (filtered by given categories to skip)
+        return [...new Set(Object.values(p_state.columnToCategoryMapping).filter(x => !p_skipCategories.has(x)))];
+    },
+
     getMappedColumns: (p_state) => (p_category) => {
 
         const mappedColumns = [];
--- a/store/index.js
+++ b/store/index.js
@@ -399,16 +399,22 @@ export const getters = {
 
                 // 1. Determine if at least one column has been linked to a category
                 const categorizationStatus = Object.values(p_state.columnToCategoryMapping)
-                                                   .some(category =>  null !== category );
+                                                   .some(category => null !== category );
 
                 // 2. Make sure one (and only one) column has been categorized as 'Subject ID'
                 const singleSubjectIDColumn = ( 1 === Object.values(p_state.columnToCategoryMapping)
                                                             .filter(category => "Subject ID" === category)
                                                             .length );
 
+                const notOnlySubjectIDCategorized = ( Object.values(p_state.columnToCategoryMapping)
+                                                            .filter(category => "Subject ID" !== category &&
+                                                                    null !== category)
+                                                            .length >= 1 );
+
                 // Annotation page is only accessible if at least one column has
                 // been categorized and if one (and only one) column has been categorized as 'Subject ID'
-                pageAccessible = categorizationStatus && singleSubjectIDColumn;
+                // and if at least one category other than Subject ID has been categorized
+                pageAccessible = categorizationStatus && singleSubjectIDColumn && notOnlySubjectIDCategorized;
 
                 break;
             }
--- a/store/index.js
+++ b/store/index.js
@@ -480,12 +480,22 @@ export const mutations = {
 
         if ( markAsMissing ) {
 
+            // 1. Create missing value list if it does not yet exist
+            // NOTE: The idea here is to only have missing value lists for columns as needed
+            // And this will be reflected in the data dictionary output from the download page
+            if ( !("missingValues" in p_state.dataDictionary.annotated[column]) ) {
+
+                p_state.dataDictionary.annotated[column].missingValues = [];
+            }
+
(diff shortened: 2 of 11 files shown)


jarmoza:
Okay @rmanaem, test fixes are up. Ready to review.

jarmoza:
Okay @rmanaem. Seb's PR is merged in and tests are green.

store/index.js
```diff
@@ -390,16 +407,22 @@ export const getters = {
...
+                const notOnlySubjectIDCategorized = ( Object.values(p_state.columnToCategoryMapping)
+                                                            .filter(category => "Subject ID" !== category &&
+                                                                    null !== category)
+                                                            .length >= 1 );
+
                 // Annotation page is only accessible if at least one column has
                 // been categorized and if one (and only one) column has been categorized as 'Subject ID'
-                pageAccessible = categorizationStatus && singleSubjectIDColumn;
+                // and if at least one category other than Subject ID has been categorized
+                pageAccessible = categorizationStatus && singleSubjectIDColumn && notOnlySubjectIDCategorized;
```
rmanaem, line 425: Is there a case where `singleSubjectIDColumn` and `notOnlySubjectIDCategorized` are `true` but `categorizationStaus` is `false`?

store/index.js
```diff
@@ -245,6 +245,19 @@ export const getters = {
...
+            .filter(category => null !== category && !p_categorySkipList.includes(category));
+
+        // 2. Create a set of the unique mapped categories
+        const categorySet = new Set(currentCategories);
+
+        // Return the categories in array form
+        return [...categorySet];
+    },
+
     getMappedColumns: (p_state) => (p_category) => {
```
rmanaem, line 261: > Because we have not yet had a discussion on how we want optional data dictionary key value pairs to be created, it is currently necessary to check for the existence of the missingValues list in both the getMappedColumns getter and changeMissingStatus mutation.

Referencing the PR description here, there is a check for `missingValues` in `changeMissingStatus` and `getUniqueValues` but not in `getMappedColumns`.

store/index.js
jarmoza: Good catch. With the new page access criteria, `categorizationStatus` can now be removed.

cypress/e2e/page/categorization-pagetests.cy.js
```diff
@@ -69,7 +69,7 @@ describe("Tests on categorization page ui via programmatic state loading and sto
                     cy.categorizeColumn("Subject ID", p_dataset["category_columns"]["Subject ID"][0]);
 
                     // 3. Assert that annotation nav and next page button are enabled
-                    cy.assertNextPageAccess("annotation", true);
+                    cy.assertNextPageAccess("annotation", false);
```
rmanaem, line 72: I see. Since we changed the requirements, the user can no longer go to the next page with single categorization.

cypress/e2e/page/categorization-pagetests.cy.js
```diff
@@ -69,7 +69,7 @@ describe("Tests on categorization page ui via programmatic state loading and sto
                     cy.categorizeColumn("Subject ID", p_dataset["category_columns"]["Subject ID"][0]);
 
                     // 3. Assert that annotation nav and next page button are enabled
```
rmanaem, line 71: not enabled

rmanaem (commented):
Thanks for the well written PR description @jarmoza!
I added some in-line comments. Aside from that I was wondering why you decided to change component tests to use `beforeEach`.

store/index.js
rmanaem: The PR description mentions a check for `missingValues` in `getMappedColumns` but there isn't one instead there is a check in `getUniqueValues`.

jarmoza: Ah! This is a typo. I'll amend it. It should read `getUniqueValues`. This refers to https://github.com/neurobagel/annotation_tool/pull/375/commits/83bd5114335d1e05a8ddf68b958a6d8dd5927ba6.

rmanaem:
> You're welcome! I thought it would be helpful given the amount of changes across files.

It really i!

>  have noticed that when Cyoress begins a new individual test that the functions defined in store.getters in particular were being read as "undefined." My workaround for this was to reinitialize the store in beforeEach before each individual test. JSON.stringify does not correctly output values assigned to an object key if they are function definitions, making the issue more difficult to debug. Not happy that I don't understand why this is occurring, but there is at least this beforeEach workaround which is not the worst thing in the world to have to do.

Yea, I think we talked about this too but this only affects the unit test, not component tests.

store/index.js
rmanaem: Thanks for clarifying that!

jarmoza:
@rmanaem 
>Yea, I think we talked about this too but this only affects the unit test, not component tests.

This is in reference to `annot-categorical.cy.js` in particular? I just reverted to defining the `store` object above the `describe` once and it works. Shall I commit the change?

rmanaem:
> This is in reference to annot-categorical.cy.js in particular?

I was referring to component tests in general. I was curious why you decided to use this syntax instead of the old one. 

> I just reverted to defining the store object above the describe once and it works. Shall I commit the change?

I leave that up to you but other than that this PR is good to go! 👨🏼‍🍳

rmanaem (approved):

jarmoza:
@rmanaem I think let's save that for another PR. I'll make an issue to look over the tests for consistency in setup.

Issue #377, opened 2023-03-15
Phase 2 refactor: Adding getSelectedOption getter


jarmoza:
This previously unspecified getter retrieves the annotated value of a raw value in a particular column. Currently used by the `annot-categorical` component on its `vue-select`. These values are found in the store's `dataDictionary.annotated.<columnName>.valueMap` object.

In addition to the implementation of this getter with this task, I am going to go ahead and propose a name change to `getCategoricalSelectedOption` for specificity sake.

@surchs Let me know if you have any comment on the functionality or name change.

surchs:
@jarmoza, how about `getSelectedCategoricalOption` as a name, that reads more clearly to me.

> getter retrieves the annotated value of a raw value in a particular column

That sounds perfect. I just want to point out that at the moment the (not yet existing) getter is being called with the row index of the table that is being displayed so that would have to be changed. I absolutely agree that the getter should accept the column and raw value, and not the row index.

jarmoza:
@surchs Sounds good re: the name. 

>I just want to point out that at the moment the (not yet existing) getter is being called with the row index of the table that is >being displayed so that would have to be changed. I absolutely agree that the getter should accept the column and raw >value, and not the row index.

That's exactly how I've implemented it. PR coming up shortly.

jarmoza:
Closed by #378

PR #380, opened 2023-03-16
Replace remaining getSelectedOption instances and fix optional chaining bug with new getter

jarmoza:
There were two remaining instances of `getSelectedOption` after PR #377 that were breaking the app when manually running through to the annotation page. (They would have been noticed with the integration test for the annotation page enabled.)

In addition, after that a bug was discovered in the improper use of the optional chaining `?.` syntax in the new `getSelectedCategorical` getter. The operator before array `[]` syntax is still `?.`

### DIFF
--- a/store/index.js
+++ b/store/index.js
@@ -146,7 +146,7 @@ export const getters = {
 
     getSelectedCategoricalOption: (p_state) => (p_column, p_rawValue) => {
 
-        return p_state.dataDictionary.annotated[p_column]?.valueMap[p_rawValue] ?? "";
+        return p_state.dataDictionary.annotated?.[p_column]?.valueMap?.[p_rawValue] ?? "";
     },
 
     getCategoryNames (p_state) {
--- a/components/annot-categorical.vue
+++ b/components/annot-categorical.vue
@@ -98,7 +98,7 @@
             ...mapGetters([
 
                 "getCategoricalOptions",
-                "getSelectedOption",
+                "getSelectedCategoricalOption",
                 "getUniqueValues",
                 "getValueDescription"
             ]),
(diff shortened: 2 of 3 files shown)


store/index.js
```diff
@@ -146,7 +146,7 @@ export const getters = {
 
     getSelectedCategoricalOption: (p_state) => (p_column, p_rawValue) => {
 
-        return p_state.dataDictionary.annotated[p_column]?.valueMap[p_rawValue] ?? "";
```
rmanaem, line 149: I'm curious, what was the behavior of the bug? Would the expression not short circuit for every null/undefined chain option (except the first one) and instead throw an error if a property didn't exist?

jarmoza: It's a good question especially if we want to use these operators in the future. It was looking for `p_rawValue` and stating that it could not find the string in `p_rawValue` in `valueMap` – which is confusing because we expect the nullish coalescing operator (`??`) to resolve anything `undefined` and return with value on the right (empty string).

Old version:
`return p_state.dataDictionary.annotated[p_column]?.valueMap[p_rawValue] ?? "";`

Here it assumes that `p_column` will be found in `annotated` and `p_rawValue` will be found in `valueMap`.

New version:
`return p_state.dataDictionary.annotated?.[p_column]?.valueMap?.[p_rawValue] ?? "";`

We need to add the `?.` operator before the array index syntax to get it to resolve appropriately.

(The extra `?.` in front of `[p_column]` is done here out of caution. In fact, it could be removed. We should technically never get here without the string in `p_column` as a key in `annotated` – but if we did it would resolved as `undefined` appropriately and hit the nullish coalescing operator and return the empty string.)

Issue #352, opened 2023-02-08
Rename store field 'columnToCategoryMapping' to 'columnToCategoryMap'


jarmoza:

jarmoza:
Closed by #383

Issue #350, opened 2023-02-07
Remove tool group component file


jarmoza:
Now that assessment tool categories will be dynamically created and automatically grouped (per category), we no longer need the `annot-tool-group` component.

It's possible that some of its code may be reused for the new functionality.

This task is to remove the old component file `annot-tool-group.vue` from the annotation tool repo once [the new feature](https://github.com/neurobagel/annotation_tool/issues/228) is in place - see #228 .

jarmoza:
Closed by #383

Issue #344, opened 2023-02-03
Remove old store file post-refactor


jarmoza:
During the refactor I wanted the old store version to stick around as an example in case we missed something during our design of the new store.

Once the store refactor is done (ostensibly with Phase 2), the previous store, now named `index-old.js` should be removed from the repo.

surchs:
@jarmoza: is this blocked? If so, could you move it to "tracked" in the new status?

jarmoza:
This is "Next", so I marked it as "Specify - Done".

jarmoza:
Closed by #383

Issue #354, opened 2023-02-09
is there something like isort for js?


surchs:
I notice we sometimes have diffs to re-sort object keys and such things. Is there maybe some tool we can configure to go through the whole codebase once to make the appropriate changes, and to then keep on top of this - e.g. as part of a pre-commit check?

jarmoza:
I believe we can do this by adding the ["sort-keys" rule](https://eslint.org/docs/latest/rules/sort-keys) to the eslint rc file, and then by installing [this plugin](https://www.npmjs.com/package/eslint-plugin-sort-keys).

surchs:
Cool! I'm in favour of adopting the eslint rule to keep things consistent. I'm less clear about the tool, it looks not very widely adopted / maintained. I could imagine that having the rule enabled and configured to show up in the IDE (and during pre-commit) might be sufficient here. wdyt?

jarmoza:
'sort-keys' definitely works. I tried it out before I commented earlier. I can see it being annoying to the coder to do it manually. I'm alright with it though myself.

That said, I don't know if `eslint --fix` will alphabetize it. If so, then that's even better.

surchs:
Yes there is something like `isort` in js, it's called https://www.npmjs.com/package/eslint-plugin-import . I'll make an issue to implement

PR #383, opened 2023-03-22
Old tool group component removal, old store removal, and store field rename

jarmoza:
This PR takes care of #350, #352, and #344 . This is mostly clean up post phase 2 store refactor.

Tool group-like functionality will be restored via custom categories, but for now its unused components are removed from the repo. The old store file was in place for reference and can now be removed. And the `columnToCategoryMapping` store field is renamed back to `columnToCategoryMap` - just out of preference and precedent.


rmanaem (commented):
Everything looks good. One thing I noticed in the description is that you mention this PR addresses #354 but there wasn't a change in the `eslintrc.js`. Perhaps a typo?

PR #389, opened 2023-03-30
Implemented `updateAnnotationCount` mutation

### LINKED ISSUE #388: Implement `updateAnnotationCount` mutation
`updateAnnotationCount` mutation is called along with `setHeuristic` and `setCategoricalOption` mutations, whenever annotation is made.
This mutation loops over the `dataDictionary.annotated` and counts the number of columns that have been annotated and updates `annotationCount`.
- A categorical column is considered annotated if it contains a non-null value for its `valueMap` property
- A continuous column is considered annotated if it contains a non-null value for its `heuristicTransformation` property

rmanaem:
This PR includes:

- Implementation of `updateAnnotationCount` mutation
   - A categorical column is considered annotated if:
      - It contains `valueMap`
      - Its `valueMap` is not empty
   - A continuous column is considered annotated if:
      - It contains `transformationHeuristic`
      - Its `transformationHeuristic` is not `null`
- `updateAnnotationCount` unit-test
- Modification made to `selectCategoricalOption` mutation, if the empty option from the v-select is selected `rawValue` is removed from the `valueMap`

Closes #388

### DIFF
--- a/store/index.js
+++ b/store/index.js
@@ -3,6 +3,7 @@ import { Set } from "core-js";
 import Vue from "vue";
 
 export const state = () => ({
+    annotationCount: 0,
 
     appSetting: {
 
@@ -604,5 +605,22 @@ export const mutations = {
 
         // Set a new transformation heuristic for this column
         Vue.set(p_state.dataDictionary.annotated[column], "transformationHeuristic", heuristic);
+    },
+
+    updateAnnotationCount(p_state) {
+        let count = 0;
+
+        Object.keys(p_state.dataDictionary.annotated).forEach(columnName => {
+            const column = p_state.dataDictionary.annotated[columnName];
+
+            if (column.valueMap && column.valueMap != {}) {
+                count++;
+            }
+            if (column.transformationHeuristic && column.transformationHeuristic != null) {
+                count++;
+            }
+        });
+
+        p_state.annotationCount = count;
     }
 };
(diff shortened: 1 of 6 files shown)


store/index.js
```diff
@@ -532,8 +533,14 @@ export const mutations = {
             p_state.dataDictionary.annotated[columnName].valueMap = {};
         }
 
-        // 1. Assign the option value to a raw value for this column
-        p_state.dataDictionary.annotated[columnName].valueMap[rawValue] = optionValue;
+        if (optionValue === null) {
```
jarmoza, line 536: It would be good to just add a brief comment here explaining the if/else statement

store/index.js
```diff
@@ -604,5 +611,22 @@ export const mutations = {
...
+    updateAnnotationCount(p_state) {
+        let count = 0;
+
+        Object.keys(p_state.dataDictionary.annotated).forEach(columnName => {
+            const column = p_state.dataDictionary.annotated[columnName];
+
+            if (column.valueMap && Object.keys(column.valueMap).length > 0) {
+                count++;
+            }
+            if (column.transformationHeuristic && column.transformationHeuristic !== null) {
```
jarmoza, line 625: You can turn this into an `else if`.

jarmoza (changes requested):
I made a few small change requests, but more importantly is that you probably need to adjust the `annot-categorical` and `annot-continuous-values` component tests. They are currently not passing, and possibly also explains the github Cypress hang.

Issue #316, opened 2023-01-20
Data dictionary loading requires case-sensitive keys - causing issue on landing/categorization pages


jarmoza:
With the phase 1 refactor, we have changed the data dictionary so that 1) a skeleton dictionary is loaded with every new data table, and 2) if a data dictionary is selected, it augments the current data dictionary (either blank or not) and does not add any new fields.

One issue that has arisen is that since JavaScript is a case-sensitive language, the capitalization for keys in the data dictionary (e.g. a BIDS-style data dictionary) matter. For example, 'description' does not match up with 'Description.`

1. Since the default skeleton dictionary contains an empty `'description` key for each column, any loading (remember, augmentation) of a further data dictionary with the lowercase 'd' will not overwrite that entry but instead add another (i.e. 'Description).
2. This causes the effect of the `column-linking-table` on the categorization page to not have visible descriptions.

What should be done?

Let's agree to data dictionary string normalization rules.

surchs:
Oh, that's a great point! I am actually not completely sure if BIDS requires these keys to be capitalized or if they just tend to be in the [spec](https://bids-specification.readthedocs.io/en/stable/03-modality-agnostic-files.html#participants-file) (and in practice). 

My sense is: let's support what BIDS does - so expect "Description" rather than "description". It would be great if we can validate the data dictionary with e.g. a jsonschema before we load it so we can tell the user if something is not correct (like the non-capitalized Description key).

jarmoza:
Sounds like a new task in the making – in the very least converting the current default `'description'` into `'Description'`

surchs:
Agree. Short term we should change all pre-defined keys (i.e. not user defined) to capitalized and then have a separate issue to discuss what to do about this more generally (i.e. validation, erroring when non-capitalized keys are given in the dictionary)

rmanaem:
It was decided to implement validation for the uploaded data dictionary based on [pydantic models in CLI](https://github.com/neurobagel/bagel-cli/blob/main/bagel/dictionary_models.py), see #393

PR #391, opened 2023-04-03
Capitalized references of `description` in store and updated affected unit tests

### LINKED ISSUE #384: BUG: Descriptions not showing in column linking table yet component test passing
In our [current dev branch](https://github.com/neurobagel/annotation_tool/tree/dev_components_talk_to_store_directly), the descriptions for columns from the data dictionary file are not visible in the 'description' column of the `column-linking-table` component.

Additionally, the component test for `column-linking-table` is passing meaning the test is not as robust as it should be. For this bug fix:

- [x] Descriptions from the data dictionary should be showing up in the 'description' column of the `column-linking-table`
- [ ] The component test for the `column-linking-table` should fail if 1) description text exists for a data table column and 2) it is not visible in the 'description' component column

rmanaem:
This PR includes the fix for the bug outlined in #384.
Changes include changing `description` references to `Description` in
- index.js
- store-mutation-initializeDataDictionary.cy.js
- store-getter-getColumDescription.cy.js

Closes #384

### DIFF
--- a/cypress/unit/store-getter-getColumnDescription.cy.js
+++ b/cypress/unit/store-getter-getColumnDescription.cy.js
@@ -5,7 +5,7 @@ const state = {
     dataDictionary : {
         annotated: {
             "describedColumn": {
-                description: "This is my first column"
+                Description: "This is my first column"
             },
             "lazyColumn": {}
         }
--- a/cypress/unit/store-mutation-initializeDataDictionary.cy.js
+++ b/cypress/unit/store-mutation-initializeDataDictionary.cy.js
@@ -30,19 +30,19 @@ describe("initializeDataDictionary", () => {
         // Assert
         expect(state.dataDictionary.userProvided).to.deep.equal({
             "col1": {
-                description: ""
+                Description: ""
             },
             "col2": {
-                description: ""
+                Description: ""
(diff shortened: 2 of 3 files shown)


jarmoza (changes requested):
@rmanaem Can you also write a test for the below condition from #384 ?

>The component test for the column-linking-table should fail if 1) description text exists for a data table column and 2) it is not visible in the 'description' component column

So in other words, the failure should be a passing test.

rmanaem:
@jarmoza I'm not quite sure what that test would look like since there already is a test checking if the column descriptions are correctly displayed in `column-linking-table.cy.js`. 
I think the original description of the #384 issue was written under the assumption that the bug was coming from `column-linking-table` however, the bug was found to be in the store.

jarmoza:
> Perhaps a test with mock data having description as their property instead of Description and expecting the description not to be displayed?

@rmanaem That would work given our current code. If we do some kind of forced capitalization later on, we can readdress.

jarmoza (approved):
In light of our discussion re: a new feature to enforce capitalization (see #316, #393), we shouldn't need the test above.

So, 👨‍🍳

Issue #390, opened 2023-04-01
Initialization of data type-specific, specialized data dictionary structures


jarmoza:
As the new version of the Vuex store is integrated, it has become apparent that checks for the existence of special structures (a.k.a. keys) on the annotated data dictionary object are muddying up our code. These currently include `valueMap` for categorical columns and `transformationHeuristic` for continuous values columns.

As discussed with @rmanaem, the proposed internal improvement being implemented here is to initialize those structures when columns and categories are linked and unlinked. The logic is as follows:

- When a column is linked to a new category any old specialized structure is deleted from `dataDictionary.annotated.<column-name>` and the new appropriate structure is added in `dataDictionary.annotated.<column-name>`.
- When a column is unlinked from any category its current specialized structure is deleted from `dataDictionary.annotated.<column-name>`

This is being implemented in conjunction with bug fix #386.

jarmoza:
Closed by #395

Issue #386, opened 2023-03-28
BUG: Dropdowns in annot-categorical not showing options


jarmoza:
Currently options are missing from the `annot-categorical` component dropdowns (`v-select`). The consequence is that satisfaction of annotation requirements for passthrough to the download page is currently blocked. This is likely related to recent changes using the `label` and/or `reduce` attributes.

The following warning is issued in the browser debug console:

```
[Vue warn]: Property or method "label" is not defined on the instance but referenced during render. Make sure that this property is reactive, either in the data option, or for class-based components, by initializing the property. See: https://vuejs.org/v2/guide/reactivity.html#Declaring-Reactive-Properties.

found in

---> <AnnotCategorical> at components/annot-categorical.vue
       <AnnotTab> at components/annot-tab.vue
         <BTab>
           <BTabs>
             <AnnotationPage> at pages/annotation.vue
               <Nuxt>
                 <DefaultLayout> at layouts/default.vue
                   <Root>
```

PR #395, opened 2023-04-05
Categorical Select bugfix + Rework of initialization of data dictionary columns for data types

jarmoza:
This PR addresses both #386 and #390. A list of changes made by file below.

**annot-categorical.vue**
- Removed `label` attribute for `v-select` as it was unnecessary and producing a warning. `value` is used for showing what was just selected and retrieving what had been selected upon component re-mount
- Added component function `selectOptionAndUpdate` as we would like to avoid multiline attribute code

**annot-continuous-values.vue**
- Added component function `selectHeuristic` as we would like to avoid multiline attribute code

**index.js**

_Formatting/streamlining_
- Added a comment to describe the new annotation count store field
- Small streamlining of `getColumnDescription`, `getHarmonizedPreview` implementation
- `updateAnnotationCount` - Updated this function for a clean check of data type for the column followed by the annotation condition. This way we do not have to worry about cascading conditions.

_Feature 390 - Initialization of specialized structures for data types_
- `alterColumnCategoryMapping` - this has been rewritten to allow for the addition of special structures `transformationHeuristic` and `valueMap` in `dataDictionary.annotated.<column-name>`. Determination of which should be added is dependent on the category assigned to each column in `dataDictionary.annotated.<column-name>`.
- `getHeuristic` - transformationHeuristic property will now always be present for continuous values-categorized columns

_Bug fix 386_
- `getSelectedCategoricalOption` - used to determine via lookup of stored `identifier` in `valueMap` what the label for that identifier is
- `selectCategoricalOption` - 1) Deletes values from the value map for categorical columns if the annotated value is removed (blank vue select),
and 2) Changes/adds the annotated value to the `valueMap` when an option is selected from the vue-select in `annot-categorical`
- `setHeuristic` - Either adds the heuristic from the drop down or deletes the heuristic for a column in `annot-continuous-values`

**Test updates for the fix and feature**
- `store-getter-getHeuristic.cy.js`
- `store-getter-getSelectedCategoricalOption.cy.js`
- `store-mutation-selectCategoricalOption.cy.js`
- `store-mutation-updateAnnotationCount.cy.js`

**Any string referring to data table column in the repo now is called `columnName`**
- `alterColumnCategoryMap.cy.js`
- `column-linking-table.cy.js`
- `annot-continuous-values.cy.js`
- `annot-columns.cy.js`
- `column-linking-table.vue` (In addition attributes for the `b-table` have been reordered.)
- `annot-columns.vue`

### DIFF
--- a/cypress/unit/store-mutation-selectCategoricalOption.cy.js
+++ b/cypress/unit/store-mutation-selectCategoricalOption.cy.js
@@ -10,11 +10,16 @@ describe("selectCategoricalOption mutation", () => {
 
             state: {
 
+                columnToCategoryMap: {},
+
                 dataDictionary: {
 
                     annotated: {
 
-                        column1: {}
+                        column1: {
+
+                            valueMap: {}
+                        }
                     }
                 }
             }
@@ -24,10 +29,9 @@ describe("selectCategoricalOption mutation", () => {
     it("Makes sure a value map is created for a column in the data dictionary", () => {
 
         // Act
-        mutations.selectCategoricalOption(store.state, {
-            optionValue: "https://example.org/female",
-            columnName: "column1",
-            rawValue: "F"
+        mutations.alterColumnCategoryMapping(store.state, {
+            category: "category1",
+            columnName: "column1"
         });
 
         // Assert
--- a/cypress/unit/store-mutation-selectCategoricalOption.cy.js
+++ b/cypress/unit/store-mutation-selectCategoricalOption.cy.js
@@ -26,18 +26,6 @@ describe("selectCategoricalOption mutation", () => {
         };
     });
 
-    it("Makes sure a value map is created for a column in the data dictionary", () => {
-
-        // Act
-        mutations.alterColumnCategoryMapping(store.state, {
-            category: "category1",
-            columnName: "column1"
-        });
-
-        // Assert
-        expect(store.state.dataDictionary.annotated.column1.valueMap).to.exist;
-    });
-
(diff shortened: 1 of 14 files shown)


store/index.js
```diff
@@ -462,25 +482,44 @@ export const mutations = {
      * @param {string} category Category the column should be mapped to
      * @param {string} column Column that will be mapped to the category
```
rmanaem, line 483: Update the docstring for `columnName`

rmanaem, line 483:
```suggestion
     * @param {string} columnName Column that will be mapped to the category
```

cypress/unit/store-mutation-selectCategoricalOption.cy.js
```diff
@@ -24,10 +29,9 @@ describe("selectCategoricalOption mutation", () => {
     it("Makes sure a value map is created for a column in the data dictionary", () => {
 
         // Act
-        mutations.selectCategoricalOption(store.state, {
-            optionValue: "https://example.org/female",
-            columnName: "column1",
-            rawValue: "F"
+        mutations.alterColumnCategoryMapping(store.state, {
```
rmanaem, line 32: Since this case now uses `alterColumnCategoryMapping`, I think it's better to move it to `store-mutation-alterColumnCategoryMapping.cy.js` unit-test. What do you think?

jarmoza:
@rmanaem Changes committed. Take a look.

rmanaem (approved):
👨🏼‍🍳

PR #397, opened 2023-04-06
Bumped cypress version to `12.9.0`

### LINKED ISSUE #293: bump cypress version
our cypress version is quite behind. For example, component testing has come out of beta last month: https://www.cypress.io/blog/2022/12/15/cypress-component-testing-for-developers/

We should update the cypress version. This is true for other packages as well, but we can track that problem somewhere else.

rmanaem:
Closes #293

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,7 +24,7 @@
       },
       "devDependencies": {
         "cy-verify-downloads": "^0.1.8",
-        "cypress": "^10.3.0",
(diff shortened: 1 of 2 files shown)


surchs (approved):
Looks good to me @rmanaem :cook: 

When we look at https://github.com/neurobagel/project/issues/74, we should make sure that cypress gets automatically updated in the future (as long as it doesn't break stuff)

PR #401, opened 2023-04-12
Removed `range` and `string` options from `transformationHeuristics`

### LINKED ISSUE #392: BUG: Not all available transformation heuristics function correctly.
`euro`, `range`, and `string` transform heuristics appear to be broken

- `euro` replaces commas with full stops/periods. It should be the opposite
- `range` produces `NaN` for single values. Is this correct?
- `string` produces the string `'missing value'` - Is this (still) correct?

rmanaem:
This PR includes the implementation of the fix for #392.

Changes include:

- Removal of the `range` and `string` options from `transformationHeuristics` in `index.js`
- Update of affected tests:
  - `annot-continuous-values.cy.js`
  - `getHarmonizedPreview.cy.js`
  - `getTransformOptions.cy.js`

Closes #392

### DIFF
--- a/cypress/component/annot-continuous-values.cy.js
+++ b/cypress/component/annot-continuous-values.cy.js
@@ -42,8 +42,8 @@ describe("Continuous values component", () => {
                 },
                 getTransformOptions: () => (p_activeCategory) => {
 
-                    // return ["float", "bounded", "euro", "range", "int", "string", "isoyear"];
-                    return ["", "float", "bounded", "euro", "range", "int", "string"];
+                    // return ["float", "bounded", "euro", "int", "isoyear"];
+                    return ["", "float", "bounded", "euro", "int"];
                 },
                 getUniqueValues: () => (p_activeCategory) => {
 
--- a/cypress/unit/store-getter-getHarmonizedPreview.cy.js
+++ b/cypress/unit/store-getter-getHarmonizedPreview.cy.js
@@ -111,34 +111,6 @@ describe("getHarmonizedPreview", () => {
         expect(harmonizedValue).to.equal(42.6);
     });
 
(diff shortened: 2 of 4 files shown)


cypress/component/annot-continuous-values.cy.js
```diff
@@ -42,8 +42,8 @@ describe("Continuous values component", () => {
                 },
                 getTransformOptions: () => (p_activeCategory) => {
 
-                    // return ["float", "bounded", "euro", "range", "int", "string", "isoyear"];
-                    return ["", "float", "bounded", "euro", "range", "int", "string"];
+                    // return ["float", "bounded", "euro", "int", "isoyear"];
```
surchs, line 45: I know this was already here, but would you mind finding out why `isoyear` is commented out and then adding the info via a `TODO` flag to the comment so we know when to add it back (and remove this comment)

cypress/unit/store-getter-getHarmonizedPreview.cy.js
surchs, line 1: ah, how nice. Removing things feels very good!

surchs (approved):
Looks good @rmanaem. I have one :cherries: comment that is unrelated to your PR: could you find out why `isoyear` isn't in the list of transforms, and then add that info to the comment (see inline comment).

Otherwise good to go from me :cook:

cypress/component/annot-continuous-values.cy.js
rmanaem: Found the reason in the description of #368 PR
> `isoyear` transformation is currently being left (commented) out so it can be more fully addressed in a more robust way in its own new issue.

Couldn't find an issue to keep track of it so made #402

Issue #341, opened 2023-02-02
Re-enable e2e tests that were previously failing


jarmoza:
#338 had us disable the e2e tests for the annotation and download pages in order to get useful summary test results during PR for the integration tests that were working (index and categorization).

Once the phase 2 refactor is in such a state as the annotation and download page e2e tests are passing, they should be re-enabled

jarmoza:
Blocked by #386

jarmoza:
Closed by #405

Issue #333, opened 2023-01-31
Refactoring annotation, download, and simple e2e tests


jarmoza:
Now that I am working on the second phase of our store refactor, I am beginning with understanding what is broken in the current e2e tests for the annotation and download pages - potentially the 'simple' (full app) e2e test.

This may not be fully completable until the end of the second phase of the refactor.

Linked to #332

rmanaem:
Blocked by #400 as the simple e2e test requires download page button to be accessible.

jarmoza:
Closed by #405

PR #405, opened 2023-04-17
Refactored e2e tests

### LINKED ISSUE #332: Refactor Cypress support function 'loadAppState'
Moving on to the annotation page, there is one more support function being used for the pages tests that needs to be refactored: `loadAppState`. This makes reference to several old Vuex actions and may need to be reworked for new logic/removals in the new Vuex store.

rmanaem:
This PR contains:
- Refactor of Cypress support function `loadAppState`
- Refactor of `annotation-pagtests.cy.js` 
- Refactor of `simple-e2etests.cy.js`
- Re-enabling the refactored tests in the e2e workflow

Closes #332  #333  #341


store/index.js
```diff
@@ -21,6 +21,11 @@ export const state = () => ({
             { label: "male", identifier: "bids:male" },
             { label: "female", identifier: "bids:female" },
             { label: "other", identifier: "bids:other" }
+        ],
+        "Diagnosis": [
+            { label: "Depressive disorder", identifier: "snomed:35489007"},
+            { label: "Parkinson's disease", identifier: "snomed:49049000"},
+            { label: "other", identifier: "snomed:other"}
```
surchs, line 28: very nice!

pages/annotation.vue
```diff
@@ -65,6 +65,11 @@
             };
         },
 
+        mounted() {
```
surchs, line 68: is this a leftover thing? Can we remove it?

pages/annotation.vue
```diff
@@ -74,7 +79,8 @@
 
             ...mapState([
 
-                "colorInfo"
+                "colorInfo",
+                "columnToCategoryMap"
```
surchs, line 83: when above is removed, remove this too.

cypress/support/commands.js
```diff
@@ -120,10 +103,15 @@ Cypress.Commands.add("datasetMeetsTestCriteria", (p_pageName, p_datasetConfig, p
 // Calls mutation in the Nuxt store
 Cypress.Commands.add("commitToVuexStore", (p_mutation, p_data) => {
 
+    console.log(`In commitToVuexStore for ${p_mutation} with data: ${JSON.stringify(p_data)}`);
```
surchs, line 106: This looks like a leftover debug statement.
If it is intentional, I think we can achieve the same thing more cleanly by using the vue-dev-utils in the browser

cypress/e2e/page/annotation-pagetests.cy.js
```diff
@@ -38,11 +38,17 @@ describe("tests on annotation page ui with programmatic state loading and store
                 cy.window().its("$nuxt.$router").then(router => {
 
                     router.push({ path: "/annotation" });
+                    cy.commitToVuexStore("setCurrentPage", "annotation");
                 });
             });
 
             it("Annotate age column; default age format transformations", () => {
 
+                console.log("Start of test store:");
```
surchs, line 47: What is the purpose of this section? If I understand right, it prints the contents of the store to console. Unless I misunderstand this should have no impact on the running of the tests, so let's remove it.

cypress/e2e/page/annotation-pagetests.cy.js
```diff
@@ -38,11 +38,17 @@ describe("tests on annotation page ui with programmatic state loading and store
                 cy.window().its("$nuxt.$router").then(router => {
 
                     router.push({ path: "/annotation" });
+                    cy.commitToVuexStore("setCurrentPage", "annotation");
```
surchs, line 41: Do we have to set `setCurrentPage` explicitly here because we're navigating to it in an atypical way for "cypress reasons"? i.e. we wouldn't have to do this during normal operations because some other handler already does the `setCurrentPage` setting for us automatically?

If so, then I think this line needs a comment that explains this fact. If the NOTE above refers to this, then I may have missed it.

cypress/e2e/page/annotation-pagetests.cy.js
```diff
@@ -58,10 +64,12 @@ describe("tests on annotation page ui with programmatic state loading and store
                     // 1. Load the app with test criteria using the dataset
                     cy.loadAppState("annotation", p_dataset, testCriteria);
 
+                    // cy.visit("/annotation");
```
surchs, line 67: Why is this line still here? If it is meant as a reminder to re-enable once some other condition is met, then we should add a comment (e.g. TODO) that says this explicitly. That said, we have git for keeping track of old code, and keeping old code in comments for longer is not a good practice. A compromise we've done in the past is to make an issue to track the future re-implementation. That is probably the right solution here too.

cypress/e2e/page/annotation-pagetests.cy.js
```diff
@@ -73,9 +81,9 @@ describe("tests on annotation page ui with programmatic state loading and store
                         .contains("li", "Age")
                         .click();
 
-                    // B. Click on the 'Save Annotation' button
-                    cy.get("[data-cy='save-button-Age']")
-                        .click();
+                    // B. Select the 'float' transformation heuristic
+                    // :data-cy="'selectTransform_' + columnName"
+                    cy.get("[data-cy='selectTransform_age']").click().type("float{enter}");
```
surchs, line 86: neato

cypress/e2e/app/simple-e2etest.cy.js
```diff
@@ -37,30 +37,32 @@ describe("End to end test using a simple UI path through the app", () => {
...
-                cy.get("[data-cy='data-dictionary-selector']")
-                    .contains("Choose file")
-                    .click()
-                    .selectFile(dataFolder + p_dataset.data_dictionary);
+            // E. Click the next page button to proceed to the categorization page
+            cy.nextPageByButton();
 
-                // E. Click the next page button to proceed to the categorization page
-                cy.nextPageByButton();
+            cy.commitToVuexStore("setCurrentPage", "categorization");
```
surchs, line 63: mentioned this for the `annotation-pagetest` as well: if this line is needed, there should be a comment to explain why. Or, if we keep re-using this pattern, we might have to add a central README where we explain this. 

For example, here we're going to the next page via the helperFunction `nextPageByButton`, and still have to explicitly set the `setCurrentPage` mutation. In the annotation page test, we use `router.push`. I'm not sure if both paths are the same under the hood (and ideally I don't have to go and read the helper function to understand the test) so a comment is probably warranted.

Before you add it, just check that we do in fact have to set this line here, because if we can remove it, we'll save two lines :)

cypress/e2e/app/simple-e2etest.cy.js
```diff
@@ -70,44 +72,50 @@ describe("End to end test using a simple UI path through the app", () => {
                 // B. Categorize "participant_id" as "Subject ID"
                 cy.categorizeColumn("Subject ID", p_dataset["category_columns"]["Subject ID"][0]);
 
-                // C. Assert nav and next button are enabled
-                cy.assertNextPageAccess("annotation", true);
+                // C. Assert nav and next button are not yet enabled
```
surchs, line 75: Could you use this comment to explain **why** they shouldn't be enabled yet? I can read the code in the next line to understand that this is **what** is being asserted, but I don't know why I would expect that assertion to be true. 

I'm assuming it is because the app doesn't consider just labeling participant and/or session as sufficient for an annotation, i.e. they are required, but not sufficient.

cypress/e2e/app/simple-e2etest.cy.js
```diff
@@ -70,44 +72,50 @@ describe("End to end test using a simple UI path through the app", () => {
...
+                cy.assertNextPageAccess("annotation", false);
 
                 // D. Categorize "age" as "Age"
                 cy.categorizeColumn("Age", p_dataset["category_columns"]["Age"][0]);
 
+                cy.assertNextPageAccess("annotation", true);
```
surchs, line 81: same as above: I would now expect a hint on why we want the assertion to be true.

cypress/e2e/app/simple-e2etest.cy.js
```diff
@@ -70,44 +72,50 @@ describe("End to end test using a simple UI path through the app", () => {
...
-                cy.assertNextPageAccess("download", true);
+                    // B. Click on the 'Age' tab
+                    cy.get("[data-cy='annotation-category-tabs'] ul")
+                        .contains("li", "Age")
+                        .click();
 
-                // E. Click the next page button to proceed to the download page
-                cy.nextPageByButton();
+                    // B. Select the 'float' transformation heuristic
+                    // :data-cy="'selectTransform_' + columnName"
```
surchs, line 101: nice, this line is helpful

surchs: you might even spell out the full sentence: "and `columnName` is `age`" or something like that to make it even clearer

surchs (changes requested):
Very cool! Congrats @rmanaem and @jarmoza on getting us back to green on the end to end tests. That's a big milestone :tada: !

I mostly have :cherries:  in the comments, but there are a couple of leftover debug statements I think you should take out. I also noted a couple of places where a clarifying comment will save a reader some time in understanding why a particular thing is done in the tests.

jarmoza:
I'm assuming @rmanaem will handle change requests here from @surchs, but if needed - given I did work on this as well - I can make requested changes for resubmission.

pages/annotation.vue
rmanaem: I think so @jarmoza can you confirm?

cypress/e2e/page/annotation-pagetests.cy.js
rmanaem: Yea, I believe it's due to Cypress reasons. I took inspiration from the `categorization-pagetests.cy.js`. I'll add a note there for clarification.

cypress/e2e/page/annotation-pagetests.cy.js
rmanaem: @jarmoza I think you added/commented this one, could you take a look, please?

cypress/e2e/app/simple-e2etest.cy.js
rmanaem: I recall setting these (in both tests) to get the right button. Without setting the current page explicitly, it seems like there is a routing/store update issue somewhere that would give the wrong current and thus the wrong nextPage e.g., the e2e test was running in the categorization page but it would get the annotation page `next-page` button. I looked at the other e2e tests namely the categorization page and found this to be the solution.
@jarmoza Did you run into this issue as well? or rather do you know why we have to explicitly set the current page?

cypress/e2e/app/simple-e2etest.cy.js
rmanaem: Yea this is due to the changes made to `isPageAccessible` getter. I'll add a comment.

cypress/e2e/app/simple-e2etest.cy.js
jarmoza: Looking into this.

cypress/e2e/page/annotation-pagetests.cy.js
jarmoza: This is a leftover line from some debugging I was doing regarding store state. It can be removed.

cypress/support/commands.js
jarmoza: All `console.log` lines in the project should be removed. They are for debug purposes only.

pages/annotation.vue
jarmoza: @rmanaem Correct. This can be removed. This was just for debugging the state of the `columnToCategoryMap` on annotation page mount.

cypress/e2e/app/simple-e2etest.cy.js
jarmoza: @surchs and @rmanaem I think I could use a second eye on this and/or will take a look with a fresh set of eyes tomorrow.

I've debugged down to some unusual behavior on the part of Cypress. In the below screen shot is from the Chrome debugger. I've commented out anything after moving to the categorization page in test and put in print statements as the test proceeds. Most importantly, I've commented out the `commitToVuexStore` call (what we think should be redundant).

Screen Shot 2023-04-19 at 5 35 54 PM

What we see is the function proceeding as normally and then once the end of the test is reached, the asynchronous calls are made – namely the call to `nextPageByButton` call.

I noticed that `setCurrentPage` is being called *twice* afterward. To check where that's happening, I wrapped the two places where `setCurrentPage` is called. Above you can see the next page button being identified as where `setCurrentPage` is being called from.

Unless I am missing something this seems like Cypress is firing the click event on the next page button twice.

rmanaem: This is consistent with what I saw during refactoring.

jarmoza: Taking another look at this morning.

jarmoza: The discussed `next-page` component test will be placed in a separate PR. It's almost ready, but regardless will not change the outcome here. We have this double-click issue with Cypress occurring, and have decieded to keep the `commitToVueXStore` calls that prevent the side effect of the test-based, extra click event from affecting the outcome of the test.

jarmoza:
@surchs / @rmanaem I will be creating a new branch for the `next-page` component tests as it turned out not to be relevant for the issue we were facing in closing out this PR. I just submitted a commit to clean up the remaining debug messages. @surchs Can you approve for merge?

rmanaem:
@jarmoza Did you make any changes to the code since my [last commit](https://github.com/neurobagel/annotation_tool/pull/405/commits/2bc0e5a4096f315c2b86bb23ba168a30bb1c8c50)?

cypress/e2e/app/simple-e2etest.cy.js
```diff
@@ -70,44 +72,54 @@ describe("End to end test using a simple UI path through the app", () => {
...
+                // C. Assert nav and next button are not yet enabled
+                // For annotation page to be enabled, subject ID and
+                // at least one other column must be categorized.
+                cy.assertNextPageAccess("annotation", false);
 
                 // D. Categorize "age" as "Age"
                 cy.categorizeColumn("Age", p_dataset["category_columns"]["Age"][0]);
 
+                // Since Age and subject ID have been categorized
+                // annotation page is no accessible.
```
surchs, line 84:
```suggestion
                // annotation page is now accessible.
```

cypress/e2e/page/annotation-pagetests.cy.js
```diff
@@ -161,192 +166,184 @@ describe("tests on annotation page ui with programmatic state loading and store
...
+                    cy.get("[data-cy='categoricalSelector_1']").click().type("Parkins{enter}");
+                    cy.get("[data-cy='categoricalSelector_2']").click().type("Other{enter}");
 
                     // 5. Assert annotation nav and next page button are enabled
                     cy.assertNextPageAccess("download", true);
                 }
             });
 
-            it("Annotate single tool group with single assessment tool column; no missing values", () => {
+            // it("Annotate single tool group with single assessment tool column; no missing values", () => {
```
surchs, line 178: As mentioned before: I am against keeping code around in a commented out way. We have it in the git log, so we can always retrieve it when we need it. There can be exceptions, but they should be rare and only used if you get to whatever is commented out next - even then there are better ways to park the code (other branch and pick, git stash if local only, ...). But at a minimum you should provide a TODO comment with a clear explanation why this is commented out and when it can be uncommented.

In this specific case, I think a better solution would be to use `it.skip(...)` to skip the test, and add the TODO comment to explain why it is skipped - but I still think not having the test until we need it is probably better.

surchs (approved):
Hey @rmanaem and @jarmoza. Looks good. I left a suggestion and a request for either removing or explicitly skipping instead of commenting out the page tests.

cypress/e2e/page/annotation-pagetests.cy.js
rmanaem: @jarmoza I'll leave this decision up to you.

jarmoza: Oh no. Let's delete this for sure! Thanks for catching that @surchs!
I'll make the change now.

PR #413, opened 2023-04-25
Next page component test

### LINKED ISSUE #412: Component tests for 'next page' button
It was revealed during the recent refactor of our e2e tests that there was no component test made for the 'next page' button. 

Since during the our e2e testing, Cypress was firing two click events from the click of the next page button, we discussed writing a component test for it to see if we could learn if this happened in the component testing environment as well. That does not appear to be the case, however the following tests were decided upon by @surchs, @rmanaem, and myself for the `next-page` component while discussing the above Cypress issue:

* Does the label on next page button correspond to the current page (This would be tested across `home`, `categorization`, and `annotation` pages – where a 'next page' button exists.)
* Does next page disabled status correspond to page accessibility
* When clicked is setCurrentPage mutation fired *once* with the correct parameters

jarmoza:
This PR closes #412.

As mentioned in the issue, the following tests have now been implemented for the `next-page` component:

* Does the label on next page button correspond to the current page (This is tested across `home`, `categorization`, and `annotation` pages – where a 'next page' button exists.)
* Does next page disabled status correspond to page accessibility
* When clicked is setCurrentPage mutation fired *once* with the correct parameters

### DIFF
--- a/.github/workflows/test_e2e.yml
+++ b/.github/workflows/test_e2e.yml
@@ -23,6 +23,7 @@ jobs:
           spec: |
             cypress/e2e/page/index-pagetests.cy.js
             cypress/e2e/page/categorization-pagetests.cy.js
+            cypress/e2e/page/annotation-pagetests.cy.js
           build: npm run build
           start: npm start
           component: false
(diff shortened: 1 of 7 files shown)


jarmoza:
@surchs Moving this to draft. I branched this off of the e2e refactor test and obviously there are some issues remaining from that misalignment we saw yesterday. To be clear, the only file changed in this PR should be `next-page.cy.js`.

PR #414, opened 2023-04-26
Next page button component tests - Clean branch

jarmoza:
(cherry picked from commit 443f742333fb34dea1c107f4d86fa215ab4bbc19) Remaking this PR off a clean branch from [dev](https://github.com/neurobagel/annotation_tool/tree/dev_components_talk_to_store_directly). 

This PR closes https://github.com/neurobagel/annotation_tool/issues/412.

As mentioned in the issue, the following tests have now been implemented for the next-page component:

* Does the label on next page button correspond to the current page (This is tested across home, categorization, and annotation pages – where a 'next page' button exists.)
* Does next page disabled status correspond to page accessibility
* When clicked is setCurrentPage mutation fired once with the correct parameters

### DIFF
--- a/cypress/component/next-page.cy.js
+++ b/cypress/component/next-page.cy.js
@@ -19,14 +19,11 @@ describe("next page button", () => {
         // Setup
         store = {
 
-            // commit: (p_mutationName, p_argument) => { store.mutations[p_mutationName](p_argument); },
             commit: () => {},
 
             getters: {
 
                 getNextPage: () => {
-
-                    // console.log(`getNextPage called with store.state: ${JSON.stringify(store.state)}`);
 
                     let nextPage = "";
 
@@ -46,14 +43,10 @@ describe("next page button", () => {
                             break;
                     }
 
-                    // console.log(`Next page is ${nextPage}`);
-
                     return nextPage;
                 },
 
                 isPageAccessible: () => (p_pageName) => {
-
-                    // console.log(`isPageAccessible called for ${p_pageName}`);
 
                     let pageAccessible = false;
 
@@ -66,8 +59,6 @@ describe("next page button", () => {
                             break;
 
                         case "categorization":
-
-                            // console.log(`dataTable: ${store.state.dataTable}`);
 
                             // Categorization page is accessible if a data table has been uploaded
                             pageAccessible = store.state.dataTable.length > 0;
@@ -102,8 +93,6 @@ describe("next page button", () => {
                             break;
                     }
 
-                    // console.log(`${p_pageName} ${(pageAccessible)? "is" : "is not"} accessible`);
-
                     return pageAccessible;
                 }
             },
@@ -111,8 +100,6 @@ describe("next page button", () => {
             mutations: {
 
                 setCurrentPage: () => (p_pageName) => {
-
-                    console.log(`setCurrentPage called for ${p_pageName}`);
 
                     store.state.currentPage = p_pageName;
                 }
--- a/cypress/component/next-page.cy.js
+++ b/cypress/component/next-page.cy.js
@@ -7,6 +7,9 @@ const uiText = {
     instructions: { "mypage": "This is how to get to the next page..." }
 };
 
+// NOTE: Due to new, cross-origin security measures in Cypress 12 for component tests,
+// testing that the click event has been fired for the next page button is blocked.
+
 describe("next page button", () => {
 
     beforeEach(() => {
@@ -20,13 +23,18 @@ describe("next page button", () => {
 
                 getNextPage: () => {
 
-                    return ( "mypage" === store.state.currentPage ) ? "mynextpage" : "mypage";
+                    console.log("getNextPage fired");
+
+                    // return ( "mypage" === store.state.currentPage ) ? "mynextpage" : "mypage";
+                    return "mynextpage";
                 }
             },
 
             mutations: {
 
                 setCurrentPage: () => (p_pageName) => {
+
+                    console.log("setCurrentPage fired");
 
                     store.state.currentPage = p_pageName;
                 }
@@ -37,8 +45,6 @@ describe("next page button", () => {
                 currentPage: "mypage",
                 dataTable: [],
                 pageData: {
-
-
 
                     mypage: {
 
@@ -142,7 +148,7 @@ describe("next page button", () => {
         cy.get("[data-cy='button-nextpage']").should("contain", uiText.button[store.state.currentPage]);
     });
 
-    it("When enabled, next page button moves to next page's url and sets current page", () => {
+    it("When enabled, next page button moves to next page's url", () => {
 
         // Setup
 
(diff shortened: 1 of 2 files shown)


surchs:
Sorry, don't have the time. Asking @rmanaem to take over!

rmanaem (approved):
Everything looks good! One thing that @surchs wanted to know and I'd like to mention is why did you use the store getters and mutations instead of mocking them in the test?

surchs:
Thanks @rmanaem for taking over. Like I said in the call, my question is whether we need to mock the getters by copying the actual logic into the test. I think one alternative is to just write the output that should get fed to the component as a literal, making the test clearer to understand. The other would be to reuse the getters in the store and then do more of an integration test. 

I prefer the former, the latter also works. But the problem I see with copying the whole logic here is that from reading the test I don't get a clear idea of what data goes into the component (missing out on clarity), and I also don't know if I'm in sync with the store (because the copy might have fallen out of sync).

jarmoza:
Understood @surchs and @rmanaem . The confusion here is that the next page component is a little bit different in that testing it could act like an integration test since its actual functionality is linked to the app state.

However, this should be tested in our 'e2e' page tests and full e2e test.

I'm revising the setup and tests now to be a proper component test for the `next-page` button.

jarmoza:
Given the above comments from @surchs and @rmanaem, I opted to convert the tests away from integration-style and more to be component specific. This PR now includes tests for the `next-page` component for the following scenarios:

*Does instruction text appear above the next page button when the button is disabled
*Button is enabled when next page is accessible and vice-versa
*Button is disabled when next page is not accessible
*Does the label on the next page button correspond to the text for the current page?
*When enabled, next page button moves to next page's url and sets current page in the store

In addition, a small bug was noticed and this has been fixed in a separate commit. Instructions text that was to appear above the `next-page` button now appears when the next page the button leads to is _inaccessible_. Previously it was checking the current page's accessibility. The idea behind the instructions text, however, is to give users instructions as to how to enable the next page button and to proceed.

jarmoza:
Just to note that this is currently failing the last component test on github only and not Cypress locally.

```
1) next page button
       When enabled, next page button moves to next page's url and sets current page:
     TypeError: Cannot set property message of  which has only a getter
      at modifyErrMsg (http://localhost:8080/__cypress/runner/cypress_runner.js:154708:15)
      at $Cy.retry (http://localhost:8080/__cypress/runner/cypress_runner.js:146130:29)
      at onFailFn (http://localhost:8080/__cypress/runner/cypress_runner.js:131085:21)
      at tryCatcher (http://localhost:8080/__cypress/runner/cypress_runner.js:8914:23)
      at Promise._settlePromiseFromHandler (http://localhost:8080/__cypress/runner/cypress_runner.js:6849:31)
      at Promise._settlePromise (http://localhost:8080/__cypress/runner/cypress_runner.js:6906:18)
      at Promise._settlePromise0 (http://localhost:8080/__cypress/runner/cypress_runner.js:6951:10)
      at Promise._settlePromises (http://localhost:8080/__cypress/runner/cypress_runner.js:7027:18)
      at _drainQueueStep (http://localhost:8080/__cypress/runner/cypress_runner.js:3621:12)
      at _drainQueue (http://localhost:8080/__cypress/runner/cypress_runner.js:3614:9)
      at ../../node_modules/bluebird/js/release/async.js.Async._drainQueues (http://localhost:8080/__cypress/runner/cypress_runner.js:3630:5)
      at Async.drainQueues (http://localhost:8080/__cypress/runner/cypress_runner.js:3500:14)
```
Currently investigating the cause.

jarmoza:
@rmanaem Here's a mystery.
Screen Shot 2023-05-10 at 2 52 50 PM

jarmoza:
@rmanaem What version of Cypress are you running locally?

rmanaem:
@jarmoza v12.9.0

jarmoza:
With the update to Cypress 12, new cross-origin security measures prevent testing of a component that navigates to a new page, particularly it is blocking the click event from being tested.

In this case, the click event for`next-page` is the mutation `setCurrentPage`. The final test has been amended to just check that the new requested url includes the next page's name.

surchs:
This seems to be quite interesting @jarmoza. Could you explain briefly what you have found or how Cypress is now preventing you from testing the button press when it passed earlier? I am assuming you are referring to the new addition of cross-origin testing: https://docs.cypress.io/guides/guides/cross-origin-testing ?

cypress/component/next-page.cy.js
```diff
@@ -0,0 +1,185 @@
...
+        cy.intercept("GET", store.getters.getNextPage(), req => {
+
+            // Assert
+
+            // Check if the url correctly includes the next page name
+            expect(req.url).to.contain(store.getters.getNextPage());
+        });
+
+        // 3. Set up a spy on the store commit function
+        cy.spy(store, "commit").as("commitSpy");
```
rmanaem, line 168: Since we ended up not spying on the mutation, I'd say we can get rid of this.

cypress/component/next-page.cy.js
```diff
@@ -0,0 +1,185 @@
...
+        // Setup
+        store = {
+
+            commit: () => {},
+
+            getters: {
+
+                getNextPage: () => {
+
+                    console.log("getNextPage fired");
```
rmanaem, line 26: I think we can get rid of `console.log` statements left over from debuggng

cypress/component/next-page.cy.js
```diff
@@ -0,0 +1,185 @@
...
+                    // return ( "mypage" === store.state.currentPage ) ? "mynextpage" : "mypage";
+                    return "mynextpage";
+                }
+            },
+
+            mutations: {
+
+                setCurrentPage: () => (p_pageName) => {
+
+                    console.log("setCurrentPage fired");
```
rmanaem, line 37: Same as above

jarmoza:
Okay @rmanaem. Requested changes + some more cleanup committed.

jarmoza:
> This seems to be quite interesting @jarmoza. Could you explain briefly what you have found or how Cypress is now preventing you from testing the button press when it passed earlier? I am assuming you are referring to the new addition of cross-origin testing: https://docs.cypress.io/guides/guides/cross-origin-testing ?
> 
> edit: the reason I'm asking is that the cypress error message refers to a getter being written to, so it would be important to know that this is in fact somehow linked to cross-origin requests.
> 

@surchs Yes, and this is a good reminder to write down/screen capture key error messages as we come across them. Thankfully we have one above.

I'll try to summarize what Arman and I were seeing when attempting to test the request url and click event. The general consensus we reached was that once a click on the button was made the request to change url could be intercepted, but the click event itself was never fired. Depending on the order of the calls of intercept and spying on the store `commit` function, we received the following behaviors.

1. The error `Cannot set property message of '' which has only a getter` - when the traceback of the object this is emanating from is expanded, we find this actually refers to an error message object and nothing in our app/store code.
2. Click followed by a page load timeout after 60 seconds
3. The error `Blocked a frame with origin "http://localhost:8080" from accessing a cross-origin frame...`

The reason above why I was not initially seeing these errors was because I was using Cypress 10.3 locally. Once upgraded to 12.9, these errors above came up. You would think that moving from '/' to 'nextpageurl' would have nothing to do with cross-origin handling since the origin should remain 'localhost', but as per error #3 and some [stackoverflow reading](https://stackoverflow.com/questions/69006948/getting-cannot-set-property-name-of-which-has-only-a-getter) about a similar error, it does seem to have something to do with a Cypress change between 10.3 and 12.9.

As an addendum:

The initial error on 10.3 I was experiencing was when the click tests were separated across two individual tests. Once a 'navigation' had been triggered via click, the Cypress testing environment would automatically fail the next test with an error reading something like `Failure to attachTo CSS or HTMLElement`. So the test would not even begin.

Issue #408, opened 2023-04-18
meta getter to step through each column and iteratively build the JSON object


surchs:
A getter to create the ready-for-download JSON object of all annotations

Then the "meta getter" should

- [x] iteratively step through each column in the data dictionary and
- [x] getter to determine whether a column has been categorized  -> no getter, just logic
- [x] getter to determine whether a column has been categorized as "discrete" or "continuous" (and in the future even "assessment tool")
- [x] rely on the category specific getters to create a JSON object for the respective column to
- [x] create a schema-valid representation of the annotations of this one column and
- [x] finally combine all of the individual columns into one big JSON object and
- [x] return this JSON object
- [x] unannotated information in the user provided data dictionary or from the data dictionary skeleton should also be included
- [x] the output format should be valid according to the data dictionary schema

surchs:
Depends on #407 and #406

jarmoza:
Taking over on this issue.

Issue #425, opened 2023-05-12
Swap 'Subject ID' for 'Participant ID'


jarmoza:
Proposal: Let's align terms used by the Neurobagel graph and the Annotation tool. Currently, the Annotation tool uses the term 'Subject ID' instead of 'Participant ID'.

- [ ] Find/replace all instances of 'Subject ID' for 'Participant ID'
- [ ] When the category is processed for annotated data dictionary output ensure the space is removed (as per https://www.neurobagel.org/documentation/dictionaries/#participant-identifier)

surchs:
@rmanaem I think this is an internal refactor that isn't tied to https://github.com/neurobagel/planning/issues/35 so we should remove it from there. Do you agree?

rmanaem:
> I think this is an internal refactor that isn't tied to https://github.com/neurobagel/planning/issues/35 so we should remove it from there. Do you agree?

Agreed.

PR #426, opened 2023-05-12
Getter for json output for discrete (categorical) value columns

### LINKED ISSUE #406: getter to create JSON representation of discrete column
Given a column mapped to a discrete category (e.g. "Diagnosis"), create a complete, schema compliant JSON object of all annotations we have made for this column.

Example from [our documentation](https://www.neurobagel.org/documentation/dictionaries/#diagnosis)
```json
  "group": {
    "Annotations": {
      "IsAbout": {
        "TermURL": "bg:diagnosis",
        "Label": "Diagnosis"
      },
      "Levels": {
        "PD": {
          "TermURL": "snomed:49049000",
          "Label": "Parkinson's disease"
        },
        "CTRL": {
          "TermURL": "purl:NCIT_C94342",
          "Label": "Healthy Control"
        }
      }
    }
  }
```
assume that
- the column was mapped to a discrete category (has entry in `state.columnToCategoryMapping`)
- `valueMapping` is incomplete
- all other data are BIDS or schema-compliant

TODO
- [ ] combine our annotations and existing annotations
- [ ] provide human readable labels (from hard-coded for now)
- [ ] provide unit tests

Data to grab

- `isAbout`
  - `state.columnToCategoryMap` for the "Label"
  - `state.categories[<categoryName>].identifier` for the IRI / TermURL
- `Levels` (in the annotations section)
  - `state.dataDictionary.annotated[<columnName>].valueMap` for the TermURL
  - `state.categoricalOptions[<categoryName>]` and then somehow find the right term (getter?)

jarmoza:
Closes #406  
See also #408 

- [ ] New getter introduced in store `getCategoricalJsonOutput` that takes a column name and outputs a json object with annotated values for that column that is compliant with the schema as noted in #406 
- [ ] New unit test file `store-getter-getCategoricalJsonOutput.cy.js` with tests for schema compliance and value/value formatting checks for the `'IsAbout'` and `'Levels'` sections of the json output

### DIFF
--- a/cypress/unit/store-getter-getCategoricalJsonOutput.cy.js
+++ b/cypress/unit/store-getter-getCategoricalJsonOutput.cy.js
@@ -8,7 +8,6 @@ let store = {
 
     state: {
 
-        appSetting: { termURLPrefix: "prefix" },
         categoricalOptions: {
 
             category1: [
@@ -16,6 +15,14 @@ let store = {
                 { label: "annotatedValue1", identifier: "https://example.org/option_1"},
                 { label: "annotatedValue2", identifier: "https://example.org/option_2"}
             ]
+        },
+        categories: {
+            "category1": {
+
+                componentName: "annot-categorical",
+                explanation: "This is an explanation for how to annotate category1.",
+                identifier: "nb:hasCategory1"
+            }
         },
         columnToCategoryMap: { column1: "category1" },
         dataDictionary: {
@@ -69,9 +76,7 @@ describe("getCategoricalJsonOutput", () => {
 
         // Assert - TermURL in this column's 'IsAbout' section is properly formatted
         // with prefix, colon, and capitalized term
-        expect(output.Annotations.IsAbout.TermURL).to.equal(
-            store.state.appSetting.termURLPrefix + ":"
-            + category[0].toUpperCase() + category.substring(1));
+        expect(output.Annotations.IsAbout.TermURL).to.equal("nb:hasCategory1");
     });
 
     it("Make sure 'Levels' contains correctly transformed value map data from the store's annotated data dictionary", () => {
--- a/cypress/unit/store-getter-getCategoricalJsonOutput.cy.js
+++ b/cypress/unit/store-getter-getCategoricalJsonOutput.cy.js
@@ -1,5 +1,7 @@
 import { getters } from "~/store";
 
+const category = "category1";
+const columnName = "column1";
 let store = {
 
     getters: getters,
@@ -10,18 +12,18 @@ let store = {
         categoricalOptions: {
 
             category1: [
+
                 { label: "annotatedValue1", identifier: "https://example.org/option_1"},
                 { label: "annotatedValue2", identifier: "https://example.org/option_2"}
             ]
         },
-        columnToCategoryMap: {
-
-            column1: "category1"
-        },
-
+        columnToCategoryMap: { column1: "category1" },
         dataDictionary: {
+
             annotated: {
+
                 column1: {
+
                     valueMap: { "rawValue1": "annotatedValue1" }
                 }
             }
@@ -31,9 +33,7 @@ let store = {
 
 describe("getCategoricalJsonOutput", () => {
 
-    it.skip("Make sure categorical json output is schema compliant", () => {
-
-        let columnName = "column1";
+    it("Make sure categorical json output is schema compliant", () => {
 
         // Act - Get formatted json output data for discrete value column
         const output = store.getters.getCategoricalJsonOutput(store.state)(columnName);
@@ -61,9 +61,6 @@ describe("getCategoricalJsonOutput", () => {
 
     it("Make sure 'IsAbout' fields are properly formatted", () => {
 
-        let columnName = "column1";
-        let category = "category1";
-
         // Act - Get formatted json output data for discrete value column
         const output = store.getters.getCategoricalJsonOutput(store.state)(columnName);
 
@@ -78,8 +75,6 @@ describe("getCategoricalJsonOutput", () => {
     });
 
     it("Make sure 'Levels' contains correctly transformed value map data from the store's annotated data dictionary", () => {
-
-        let columnName = "column1";
 
         // Act - Get formatted json output data for discrete value column
         const output = store.getters.getCategoricalJsonOutput(store.state)(columnName);
--- a/cypress/unit/store-getter-getCategoricalJsonOutput.cy.js
+++ b/cypress/unit/store-getter-getCategoricalJsonOutput.cy.js
@@ -31,7 +31,7 @@ let store = {
 
                 column1: {
 
-                    valueMap: { "rawValue1": "annotatedValue1" }
+                    valueMap: { "rawValue1": "https://example.org/option_1" }
                 }
             }
         }
@@ -88,12 +88,11 @@ describe("getCategoricalJsonOutput", () => {
         expect(Object.keys(output.Annotations.Levels)).to.deep.equal(
             Object.keys(store.state.dataDictionary.annotated[columnName].valueMap));
 
-        // Assert - Labels in each object in 'Levels' should match annotated value from the
+        // Assert - Labels in each object in 'Levels' should match label from the
         // store's annotated data dictionary value map
         Object.keys(output.Annotations.Levels).forEach(rawValue => {
 
-            expect(output.Annotations.Levels[rawValue].Label).to.equal(
-                store.state.dataDictionary.annotated[columnName].valueMap[rawValue]);
+            expect(output.Annotations.Levels[rawValue].Label).to.equal("annotatedValue1");
         });
     });
 });
--- a/cypress/unit/store-getter-getCategoricalJsonOutput.cy.js
+++ b/cypress/unit/store-getter-getCategoricalJsonOutput.cy.js
@@ -1,7 +1,5 @@
 import { getters } from "~/store";
 
-const category = "category1";
-const columnName = "column1";
 let store = {
 
     getters: getters,
@@ -32,7 +30,7 @@ let store = {
                 column1: {
 
                     valueMap: { "rawValue1": "https://example.org/option_1" },
-                    missingValues: {0: "missing"}
+                    missingValues: ["missing"]
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


surchs:
ah, didn't realize you had already picked this up @rmanaem.

rmanaem:
@surchs No worries

cypress/unit/store-getter-getCategoricalJsonOutput.cy.js
```diff
@@ -0,0 +1,94 @@
...
+
+        // Act - Get formatted json output data for discrete value column
+        const output = store.getters.getCategoricalJsonOutput(store.state)(columnName);
+
+        // Assert - Label in this column's 'IsAbout' section to be its assigned category
+        expect(output.Annotations.IsAbout.Label).to.equal(category);
+
+        // Assert - TermURL in this column's 'IsAbout' section is properly formatted
+        // with prefix, colon, and capitalized term
+        expect(output.Annotations.IsAbout.TermURL).to.equal(
```
rmanaem, line 72: I think it's better to be explicit in this case i.e. use `prefix:Category1` instead of the code used inside the getter for formatting the `TermURL` because if the code is not doing what its supposed to be doing the test won't pick up on it.

rmanaem (changes requested):
Thanks for the PR @jarmoza! I requested a minor change in one of the test cases, let me know what you think.

surchs:
Just mentioning a couple of things I noticed here:

- this was missing from the spec, but we have a process for "missing values" and they should be part of the output data dictionary. See https://www.neurobagel.org/documentation/dictionaries/#assessment-tool for example.
- what is [stored under `valueMap` ](https://github.com/neurobagel/annotation_tool/blob/f29544e942d11c3fc90a20316f71de4b94ae9c06/store/index.js#L575 )in the store is not the text-label of a `categoricalOption`, but the unique `identifier`. See [the relevant component test](https://github.com/neurobagel/annotation_tool/blob/f29544e942d11c3fc90a20316f71de4b94ae9c06/cypress/component/annot-categorical.cy.js#L107) for an example -> we have the identifier, and we need to lookup the human label (not the other way around)
- although it is true that currently all variables a column can be `"isAbout"` live in the neurobagel namespace, not all of them will have a unique identifier that is an exact match with their human readable label. I see @jarmoza has already made an issue to align the label with the identifier: https://github.com/neurobagel/annotation_tool/issues/425 However I think this will quickly get hard to keep up. I think a better way is to do what we've done for the (currently hardcoded) [categorical options](https://github.com/neurobagel/annotation_tool/blob/f29544e942d11c3fc90a20316f71de4b94ae9c06/store/index.js#L26) and add an `identifier` attribute to the pre-defined categories (e.g. [here](https://github.com/neurobagel/annotation_tool/blob/f29544e942d11c3fc90a20316f71de4b94ae9c06/store/index.js#L35)). That will allow us to 
  - have the identifier clearly spelled out in a single place - reducing mental math to parse it from the code
  - have a variable mapped to a different identifier than it's human label (e.g. something with a space `"Participant ID"` could map to `http://neurobagel.org/vocab/ParticipantID`)
  - change to a different, non neurobagel namespace, ...

cypress/unit/store-getter-getCategoricalJsonOutput.cy.js
```diff
@@ -0,0 +1,106 @@
+import { getters } from "~/store";
+
+const category = "category1";
+const columnName = "column1";
```
surchs, line 4: :cherries:: why do we need these variables? Don't they just refer to what's mocked? Is the idea that I could change the name of the mocked stuff and then I can change it all in one place? Not sure this is worth the reduced readability.

cypress/unit/store-getter-getCategoricalJsonOutput.cy.js
```diff
@@ -0,0 +1,106 @@
...
+        },
+        columnToCategoryMap: { column1: "category1" },
+        dataDictionary: {
+
+            annotated: {
+
+                column1: {
+
+                    valueMap: { "rawValue1": "https://example.org/option_1" },
+                    missingValues: {0: "missing"}
```
surchs, line 35:
```suggestion
                    missingValues: ["missing"]
```
`missingValues` is an array, see e.g. https://github.com/neurobagel/annotation_tool/blob/784b974636b8f72740d4ca98138f153d81e70a61/cypress/unit/store-mutation-changeMissingStatus.cy.js

cypress/unit/store-getter-getCategoricalJsonOutput.cy.js
```diff
@@ -0,0 +1,106 @@
...
+    it("Make sure categorical json output is schema compliant", () => {
+
+        // Act - Get formatted json output data for discrete value column
+        const output = store.getters.getCategoricalJsonOutput(store.state)(columnName);
+
+        // Assert - Current annotated data dictionary schema compliance
+
+        // Parent level
+
+        expect(output).to.have.property("Annotations");
```
surchs, line 53: :cherries:: I find the following section quite hard to read and understand. I think it'd be clearer to just write out the expected json and then `deep.equal` it (assuming that would work).

cypress/unit/store-getter-getCategoricalJsonOutput.cy.js
```diff
@@ -0,0 +1,106 @@
...
+        });
+    });
+
+    it("Make sure 'IsAbout' fields are properly formatted", () => {
+
+        // Act - Get formatted json output data for discrete value column
+        const output = store.getters.getCategoricalJsonOutput(store.state)(columnName);
+
+        // Assert - Label in this column's 'IsAbout' section to be its assigned category
+        expect(output.Annotations.IsAbout.Label).to.equal(category);
```
surchs, line 77: :cherries:: Agree with your previous review comment on making this explicit (which you've done below). I think it'd make sense to also replace `category` with the actual name (`"category1"`) here (see above).

cypress/unit/store-getter-getCategoricalJsonOutput.cy.js
```diff
@@ -0,0 +1,106 @@
...
+        // Act - Get formatted json output data for discrete value column
+        const output = store.getters.getCategoricalJsonOutput(store.state)(columnName);
+
+        // Assert - Keys in 'Levels' are raw values from the value map
+        expect(Object.keys(output.Annotations.Levels)).to.deep.equal(
+            Object.keys(store.state.dataDictionary.annotated[columnName].valueMap));
+
+        // Assert - Labels in each object in 'Levels' should match label from the
+        // store's annotated data dictionary value map
+        Object.keys(output.Annotations.Levels).forEach(rawValue => {
```
surchs, line 95: I find this quite confusing here. If I understand correctly, there is nothing to iterate over because there is only a single value that is mapped to anything. Is that correct? If so, let's drop this `forEach` here and be explicit and just put the expected `Annotations.Levels` object in the test. The `forEach` implies that there are multiple values - which makes the assertion that they should all be mapped to `"annotatedValue1"` misleading and also brittle if we ever add more mapped values to the test - which is what I'm assuming this construct is trying to anticipate?

cypress/unit/store-getter-getCategoricalJsonOutput.cy.js
```diff
@@ -0,0 +1,106 @@
...
+
+            expect(output.Annotations.Levels[rawValue].Label).to.equal("annotatedValue1");
```
surchs, line 97: linked to above: let's just put the actual expected object in here. Helps with readability and avoid this multi-step assertion

cypress/unit/store-getter-getCategoricalJsonOutput.cy.js
```diff
@@ -0,0 +1,106 @@
...
+        });
+    });
+
+    it("Make sure 'MissingValues' is the same as 'missingValues' from annotated data dictionary ", () => {
+        const output = store.getters.getCategoricalJsonOutput(store.state)(columnName);
+
+        expect(Object.keys(output.Annotations.MissingValues)).to.deep.equal(Object.keys(["missing"]));
```
surchs, line 104: As above: `missingValues` is not an object but it's an array. Both inside the `store.dataDictionary` and in the desired output `.json`: https://www.neurobagel.org/documentation/dictionaries/#assessment-tool

store/index.js
```diff
@@ -35,17 +35,20 @@ export const state = () => ({
         "Age": {
 
             componentName: "annot-continuous-values",
-            explanation: "This is an explanation for how to annotate age."
+            explanation: "This is an explanation for how to annotate age.",
+            identifier: "nb:hasAge"
```
surchs, line 39: The term to use here would be for the class "Age", not the property "hasAge", so `nb:Age` instead of `nb:hasAge`. Just pinging @alyssadai here quickly to see if there is a nice place where this is listed - and if there isn't, we might want to consider one?

store/index.js
```diff
@@ -146,6 +149,54 @@ export const getters = {
...
+                },
+                Levels: {},
+                MissingValues: []
+            }
+        };
+
+        // 1. Fill out Annotations 'IsAbout' section
+
+        // A. Label matches the assigned category
+        formattedOutput.Annotations.IsAbout.Label = category;
```
surchs, line 174: This is correct! I'm just thinking: now that we have an `identifier` attribute for each category, should we use it to refer to the category (like we're doing in other cases for mappings)? Definitely not for this PR, but maybe worth an issue to discuss.

store/index.js
```diff
@@ -146,6 +149,54 @@ export const getters = {
...
+
+        // B. Term matches the category identifier
+        formattedOutput.Annotations.IsAbout.TermURL = p_state.categories[category].identifier;
```
surchs, line 177: nice!

store/index.js
```diff
@@ -146,6 +149,54 @@ export const getters = {
...
+
+        // 2. Fill out Annotations 'Levels' section
+        Object.keys(annotatedDictColumn.valueMap).forEach(rawValue => {
+
+            formattedOutput.Annotations.Levels[rawValue] = {};
+
+            p_state.categoricalOptions[category].forEach(option => {
```
surchs, line 184: couldn't we just do 

```js
const myOption = p_state.categoricalOptions[category].find(option => option.identifier === annotatedDictColumn.valueMap[rawValue])
```
here? And then do the assignment in one go with the object declaration above, i.e.:

```js
formattedOutput.Annotations.Levels[rawValue] = {
"Label": myObject.label
"TermURL": myObject.identifier
}
```

store/index.js
```diff
@@ -146,6 +149,54 @@ export const getters = {
...
+
+                if ( annotatedDictColumn.valueMap[rawValue] === option.identifier ) {
+                    formattedOutput.Annotations.Levels[rawValue].Label = option.label;
+                    formattedOutput.Annotations.Levels[rawValue].TermURL = option.identifier;
+                }
+            });
+        });
+
+        Object.keys(annotatedDictColumn.missingValues).forEach(key => {
+            formattedOutput.Annotations.MissingValues.push(annotatedDictColumn.missingValues[key]);
```
surchs, line 194: I think this is a bit redundant. We should just be able to do

```js
formattedOutput.Annotations.MissingValues = annotatedDictColumn.missingValues
```

surchs (changes requested):
Thanks a lot @rmanaem for taking on the PR! I read through the whole thing and left some comments. I think the `missingValues` implementation is assuming the wrong data structure for `missingvalues` (it's an array, not an object). Could you fix that and take a look at some of the other comments (:cherries: are just nice-to-have's you can ignore).

store/index.js
alyssadai: Yep - I think this page https://www.neurobagel.org/documentation/dictionaries/#age has the up-to-date info for this variable 🙂

cypress/unit/store-getter-getCategoricalJsonOutput.cy.js
rmanaem: > why do we need these variables? 

These are left over from Jonathan's code.

>  Is the idea that I could change the name of the mocked stuff and then I can change it all in one place? 

I think so

I'll clean them up.

cypress/unit/store-getter-getCategoricalJsonOutput.cy.js
rmanaem: Something like this?
```Javascript
expect(output).to.deep.equal(
    {

        Annotations: {

            IsAbout: {

                Label: "category1",
                TermURL: "nb:hasCategory1"
            },
            Levels: {
                rawValue1: {
                    Label: "annotatedValue1",
                    TermURL: "https://example.org/option_1"
                }
            },
            MissingValues: ["missing"]
        }
})
```
This does work.

surchs: I find this much clearer, yes, because I can immediately understand what the passed data looks like.

cypress/unit/store-getter-getCategoricalJsonOutput.cy.js
rmanaem: > The forEach implies that there are multiple values - which makes the assertion that they should all be mapped to "annotatedValue1" misleading and also brittle if we ever add more mapped values to the test - which is what I'm assuming this construct is trying to anticipate?

Yea, It used to be something implicit but I changed it while I was reworking the getter and its test.
Since we're sticking with one value, I'll get rid of the `forEach`.

store/index.js
rmanaem: I think so

rmanaem (approved):

rmanaem (approved):

surchs (approved):
Nice, looks good @rmanaem. Good to merge :cook: 

Thanks @jarmoza and @rmanaem for addressing this!

Issue #427, opened 2023-05-12
Restoration of download functionality for refactor


jarmoza:
Now that we are moving toward creating the proper, BIDS-formatted output from the new refactor for the annotation tool, we also need to restore working download functionality on the download page.

#404 moved part way toward this under the frame of #398.

Work remaining for this to be fully functional (aside from the getters to get formatted output):

- [x] New store field and mutation to retain the filename of the data dictionary when it is uploaded (along with a 'default' filename when there is no data dictionary) See #416.
- [x] Updates to the download page for the new filename
- [x] Updates to the download page for the new json output getter(s)

jarmoza:
@rmanaem See my branch https://github.com/neurobagel/annotation_tool/tree/restore-download-functionality where store field/mutation and download page filename updates are addressed thus far.

jarmoza:
Blocked by #408

jarmoza:
Closed by #433

PR #431, opened 2023-05-24
[ENH] Meta-getter for JSON output from download page

### LINKED ISSUE #408: meta getter to step through each column and iteratively build the JSON object

jarmoza:
Closes #408 

Meta-getter

- New store getter `getJsonOutput` assembles into one JS object, child JS objects from categorical and continuous columns as well as other entries in the data dictionary that have not been added by the Annotation tool, but rather from users at time of upload of the data dictionary
- Unit tests for `getJsonOutput` in `store-getter-getJsonOutput.cy.js`

Helpers for the Meta-getter

- New getter `getColumnDataType` retrieves `componentName` from `categoricalOptions` when given a column name. This looks up the category that has been linked to the column, determining whether it has been assigned a categorical value-based or continuous value-based annotation. If the column has not yet been categorized, returns null
- Unit tests for `getColumnDataType` in `store-getter-getColumnDataType.cy.js`

Changes to current code
- `MissingValues` now added and checked for in the unit tests of the `getContinuousJsonOutput` getter
- Some formatting cleanup + additional comments in `index.js`


jarmoza:
@rmanaem Just a note that the calls to the getters in `getJsonOutput` may have to be amended. As I'm currently trying to use this meta getter in #427 and running into some issues getting the function to run.

Currently, in this PR the tests written for the meta getter pass, but as you can see they are called somewhat unusually by passing in the state variable and then that function call is passing the argument (`columnName` in the case of all three child getters).

jarmoza:
@rmanaem Nevermind about my comment above regarding a possible new commit. I was testing out `getJsonOutput` for #427 just now, and I believe it will function with this code as is.

store/index.js
```diff
@@ -346,6 +393,46 @@ export const getters = {
...
+
+        let jsonOutput = {};
+
+        // 1. Create a JSON output that contains annotations made thus far
+        Object.keys(p_state.dataDictionary.annotated).forEach(columnName => {
+
+            let columnOutput;
+
+            // A. If a column has been categorized/annotated, transform its annotations to json output format
+            if ( null !== p_state.columnToCategoryMap[columnName] ) {
```
rmanaem, line 406: If I'm not mistaken `columnToCategoryMap` only contains mappings made on categorization page. We allow users to map a column to a category but don't enforce annotation of every column i.e., they can go past the annotation page and proceed to download page as long as they've annotated one column. So I think checking the `columnToCategoryMap` as the only condition will result in skipping columns that have been categorized but not annotated since there is no annotation for these columns. Not sure if that's what we want, what do you think?

store/index.js
```diff
@@ -275,12 +305,29 @@ export const getters = {
...
         formattedOutput.Annotations.IsAbout.Label = category;
 
+        // B. Term matches the category identifier
         formattedOutput.Annotations.IsAbout.TermURL = p_state.categories[category].identifier;
 
+        // 2. Save the transformation heuristic from the list of possible heuristics
         formattedOutput.Annotations.Transformation = p_state.transformationHeuristics[annotatedDictColumn.transformationHeuristic];
 
+        // 3. Save the missing values
+        formattedOutput.Annotations.MissingValues = annotatedDictColumn.missingValues;
```
rmanaem, line 320: To adhere to the [exisitng documentation](https://www.neurobagel.org/documentation/dictionaries/#age), I think it's better to remove the `MissingVlaues`.

cypress/unit/store-getter-getContinousJsonOutput.cy.js
```diff
@@ -62,20 +66,22 @@ describe("getcontinuousJsonOutput", () => {
...
+        expect(output).to.deep.equal({
+
+            Annotations: {
+
+                IsAbout: {
+
+                    Label: "category1",
+                    TermURL: "nb:hasCategory1"
+                },
+                MissingValues: [],
```
rmanaem, line 78: See above.

cypress/unit/store-getter-getJsonOutput.cy.js
```diff
@@ -0,0 +1,124 @@
...
+        categories: {
+
+            category1: {
+
+                componentName: "annot-categorical",
+                identifier: "category1Identifier"
+            },
+            category2: {
+
+                componentName: "annot-continuous",
```
rmanaem, line 29: `annot-continuous-values`*

cypress/unit/store-getter-getJsonOutput.cy.js
```diff
@@ -0,0 +1,124 @@
...
+                Label: "transform1",
+                TermURL: "transform1Identifier"
+            }
+        }
+    }
+};
+
+describe("getJsonOutput", () => {
+
+    it("Output data dictionary entries for data table columns are properly formatted according to their assigned data type", () => {
```
rmanaem, line 73: Suggestion: I think having one `expect` statement that checks the entirety of the `jsonOutput` improves readability here.

cypress/unit/store-getter-getJsonOutput.cy.js
```diff
@@ -0,0 +1,124 @@
...
+        // Also includes extra user provided data dictionary definition and missing values
+        expect(jsonOutput["continuousColumn"]).to.deep.equal({
+
+            Annotations: {
+
+                IsAbout: {
+                    Label: "category2",
+                    TermURL: "category2Identifier"
+                },
+                MissingValues: ["1", "2", "3"],
```
rmanaem, line 111: Remove `MissingValues`, see previous comments.

rmanaem (changes requested):
Thanks for the PR @jarmoza! I left a few comments for requested changes and one as a suggestion. Let me know what you think.

store/index.js
```diff
@@ -346,6 +393,46 @@ export const getters = {
...
+
+                // I. Columns with different data types yield different json outputs
+                switch ( p_getters.getColumnDataType(p_state)(columnName) ) {
+
+                    case "annot-categorical":
+
+                        columnOutput = p_getters.getCategoricalJsonOutput(p_state)(columnName);
+                        break;
+
+                    case "annot-continuous":
```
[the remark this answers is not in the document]
jarmoza: Got it. The test had the wrong mock value.

store/index.js
jarmoza: The logic currently is as follows:
```
if column has been categorized:
     - include in its data dictionary entry any non-Neurobagel contents and our 'Annotations' section
else:
    - include in its data dictionary entry any non-Neurobagel contents
```
When `initializeColumnToCategoryMap` is called by `processDataTable`, it initializes the `columnToCategoryMap` with all columns as keys and values as null. The reason `columnToCategoryMap` and categorization of a column is used here is because without categorization the column will not show up on the annotation page tabs. Thus we do know what at least had the possibility of being annotated.

> So I think checking the columnToCategoryMap as the only condition will result in skipping columns that have been categorized but not annotated since there is no annotation for these columns. Not sure if that's what we want, what do you think?

Maybe what you mean is: we can have empty annotations here since only one annotated column is needed to move to the download page? If so, I agree we can probably make this function not output a blank `Levels` and `Missing Values` for unannotated columns. Another related change would be to not include `Missing Values` if it is blank. This would be similar for continuous value columns and `Transformation`

store/index.js
```diff
@@ -215,15 +222,27 @@ export const getters = {
...
                 if ( annotatedDictColumn.valueMap[rawValue] === option.identifier ) {
+
                     formattedOutput.Annotations.Levels[rawValue].Label = option.label;
                     formattedOutput.Annotations.Levels[rawValue].TermURL = option.identifier;
                 }
             });
         });
 
+        // 3. Save any other keys from data dictionary
+        Object.keys(annotatedDictColumn).forEach(columnEntry => {
```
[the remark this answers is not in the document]
jarmoza: The idea here is to preserve whatever they passed in via the user provided dictionary (if they provided their own data dictionary) on the index page. The answer is "if it's BIDS compliant to do so, then yes."

cypress/unit/store-getter-getJsonOutput.cy.js
jarmoza: Yes, I originally modeled this `expect` after your usage. But in this case, I purposely separated the cases. IMO this improves readability and more importantly the ability to separate out the tests of the differing formats across the three different case.

store/index.js
rmanaem: Thanks for the explanation, I forgot we keep empty annotations for unannotated columns. 

> Maybe what you mean is: we can have empty annotations here since only one annotated column is needed to move to the download page? If so, I agree we can probably make this function not output a blank Levels and Missing Values for unannotated, categorical columns. Another related change would be to not include MissingValues if it is blank. This would be similar for continuous value columns and Transformation and its MissingValues.

Yea, that's what I meant.

store/index.js
rmanaem: Not quite, let's ask @surchs.

store/index.js
jarmoza: Do you think the output should stay is as we should spec out changes in another issue/PR? Or make the change now not to output blank annotation entries?

rmanaem: I'd say if it can be implemented in the `getJsonOutput` it's better to handle it here but if it requires changes in other parts of the code handle it separately in another issue/PR.

rmanaem: To be discussed upon implementation of https://github.com/neurobagel/bagel-cli/issues/144

jarmoza: @rmanaem I'm holding off on merge until that issue is complete then?

store/index.js
surchs: What we discussed:
1) MissingValues is actually implemented for continuous values, it is just missing a test to show that feature -> issue #
2) Since MissingValues for continuous values is OK with the spec (and the CLI should know what to do with it), we will include MissingValues in the output of the annotator (and also test that case)
3) To prevent future issues around ambiguity of "what's in the spec" and so on, we opened issue https://github.com/neurobagel/bagel-cli/issues/144

store/index.js
rmanaem: No, I'd say let's get this out of the way and if there is any changes required we can open an issue and address them then.

jarmoza: Any further changes to this code from new documentation will be addressed in a subsequent PR.

store/index.js
jarmoza: MissingValues will remain unless schema documentation changes to state otherwise.

rmanaem (approved):
🧑‍🍳

PR #433, opened 2023-05-29
[ENH] Restore download functionality for refactor/new meta getter

### LINKED ISSUE #172: Verify download testing bug
Currently, the only assert for verifying a json file has downloaded at the end our e2e test on the download page is using `verifyDownload`. For some reason, this functionality now stalls and times out.

This needs to be resolved in addition to possibly enhancing the assert to include valid output json.

### LINKED ISSUE #416: Mutation to store the data dictionary filename
We need to store the uploaded data dictionary's filename when one is uploaded on the index page. This is so that it can be used in the filename of the outputted, newly annotated data dictionary on the download page.

The filename of an uploaded file was already being passed by the `file-selector` component, but nothing was being done with it.

- [x] Create a store field `filename` in the store's `dataDictionary` object
- [x] Create a new mutation, `setDataDictionaryFilename`, that will take a filename and store it in the above new field
- [x] This mutation will be called inside the `processDataDictionary` action

This new mutation will be used for issues #406, #407, #408.

### LINKED ISSUE #427: Restoration of download functionality for refactor

jarmoza:
Closes #427 
Closes #416
Fixes #172 

- Refactor of download page to use new meta getter `getJsonOutput`
- Adds new field and getter for `dataDictionary.filename` to (re)use for the new output json file
- Download page 'e2e' test checks if a file has has downloaded after button click (solution suggestion found here: https://stackoverflow.com/questions/73971115/how-do-i-list-the-contents-of-the-download-directory-in-cypress-i-need-to-make)
- Adds app state loading code for download page
- Update unit test for meta getter `getJsonOutput` with mocks for its getters. This overcomes an issue where calling a getter from a getter worked one way in e2e/manual testing but not in unit testing and vice versa.

### DIFF
--- a/cypress/e2e/page/download-pagetests.cy.js
+++ b/cypress/e2e/page/download-pagetests.cy.js
@@ -21,26 +21,43 @@ describe("tests on download page ui via programmatic state loading and store int
 
             beforeEach(() => {
 
+                // 1. Open index page
+                cy.visit("/");
+
+                // 2. Load test data
+                cy.loadTestDataIntoStore(p_dataset);
+
+                // 3. Move to the download page
+                cy.window().its("$nuxt.$router").then(router => {
+
+                    // A. Route to annotation page
+                    router.push({ path: "/download" });
+
+                    // B. Once routing is complete, set the current page in the nuxt store
+                    // (normally would happen via navigation clicks)
+                    cy.commitToVuexStore("setCurrentPage", "download");
+                });
+
+                // 4. Categories required for this test and the number of required columns for each category
+                const testCriteria = {
+
+                    categories: [
+
+                        ["Subject ID", 1],
+                        ["Age", 1],
+                        ["Sex", 1]
+                    ]
+                };
+
+                // 5. Load the app with test criteria using the dataset
+                cy.loadAppState("download", p_dataset, testCriteria);
             });
 
-            it("", () => {
+            it("Clicking download button downloads annotated data dictionary JSON", () => {
 
-            });
-
-            it("", () => {
-
-            });
-
-            it("", () => {
-
-            });
-
-            it("", () => {
-
-            });
-
-            it("", () => {
-
+                // 1. Download the annotated data dictionary in Neurobagel JSON format
+                cy.get("[data-cy='download-button']")
+                    .click();
             });
         });
 
--- a/cypress/e2e/page/download-pagetests.cy.js
+++ b/cypress/e2e/page/download-pagetests.cy.js
@@ -55,12 +55,20 @@ describe("tests on download page ui via programmatic state loading and store int
 
             it("Clicking download button downloads annotated data dictionary JSON", () => {
 
-                // 1. Download the annotated data dictionary in Neurobagel JSON format
-                cy.get("[data-cy='download-button']")
-                    .click();
+                // 0. Checking downloads folder state before click
+                cy.task("downloads", "cypress/downloads").then(folderStateBefore => {
+
+                    // 1. Download the annotated data dictionary in Neurobagel JSON format
+                    cy.get("[data-cy='download-button']")
+                        .click();
+
+                    // 2. Check number of files in downloads folder has increased by one
+                    cy.task("downloads", "cypress/downloads").then(folderStateAfter => {
+
+                        expect(folderStateAfter.length).to.be.eq(folderStateBefore.length + 1);
+                    });
+                });
             });
         });
-
     });
-
 });
--- a/cypress/e2e/page/download-pagetests.cy.js
+++ b/cypress/e2e/page/download-pagetests.cy.js
@@ -30,7 +30,7 @@ describe("tests on download page ui via programmatic state loading and store int
                 // 3. Move to the download page
                 cy.window().its("$nuxt.$router").then(router => {
 
-                    // A. Route to annotation page
+                    // A. Route to download page
                     router.push({ path: "/download" });
 
                     // B. Once routing is complete, set the current page in the nuxt store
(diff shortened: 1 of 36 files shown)


cypress/e2e/page/download-pagetests.cy.js
```diff
@@ -21,29 +21,54 @@ describe("tests on download page ui via programmatic state loading and store int
...
 
+                // 5. Load the app with test criteria using the dataset
+                cy.loadAppState("download", p_dataset, testCriteria);
             });
 
-            it("", () => {
+            it("Clicking download button downloads annotated data dictionary JSON", () => {
+
+                // 0. Checking downloads folder state before click
+                cy.task("downloads", "cypress/downloads").then(folderStateBefore => {
```
surchs, line 59: This too should get a section in the internal wiki. 
I also think the comment here would be more helpful if it explained why `cy.task("downloads")` works to check any download folder. I'm assuming `"downloads"` is a builtin / default funtion, and the path is cypress specific? That'd be helpful to understand the test. 

If it's in the wiki, I think a link to the wiki would also do.

cypress/e2e/page/download-pagetests.cy.js
```diff
@@ -21,29 +21,54 @@ describe("tests on download page ui via programmatic state loading and store int
...
 
+                    // 1. Download the annotated data dictionary in Neurobagel JSON format
+                    cy.get("[data-cy='download-button']")
+                        .click();
+
+                    // 2. Check number of files in downloads folder has increased by one
+                    cy.task("downloads", "cypress/downloads").then(folderStateAfter => {
+
+                        expect(folderStateAfter.length).to.be.eq(folderStateBefore.length + 1);
```
surchs, line 68: If I understand correctly, the `folderStateAfter` array will have the filename in it. Might be worth also asserting over the filename as well, since this is an e2e test (maybe without the datetime addition)?

surchs (approved):
Very exciting milestone @jarmoza :tada:. We definitely need to celebrate this big achievement with everyone, especially @rmanaem!

The PR looks good, my comments are mostly :cherries: or thoughts. Take a look and decide, then good to squash merge :cook: 

Once you do, please directly open the next PR to merge `dev_...` into `master`! We can deal with any required rebasing in that other PR then.

pages/download.vue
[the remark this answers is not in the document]
jarmoza: Got it.

pages/download.vue
```diff
@@ -80,12 +96,10 @@
...
-                // // 1. Format the annotated data dictionary into propietary JSON format
-                // const jsonData = this.transformAnnotatedDictionaryToJSON();
-
-                // // 2. Open file dialog to prompt the user to name it and
-                // // download it to their location of choice
-                // this.fileSaverSaveAs(jsonData);
+                // Open file dialog to prompt the user to name the
+                // JSON-formatted annotated data dictionary and download it to
+                // their location of choice
+                this.fileSaverSaveAs(this.getJsonOutput);
```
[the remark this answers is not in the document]
jarmoza: Nope. This wrapper is just legacy at this point. I'll place the call directly on the button.

store/index.js
```diff
@@ -643,7 +644,11 @@ export const actions = {
 
     processDataDictionary({ state, commit, getters }, { data, filename }) {
 
+        // 1. Save the user-provided data dictionary
         commit("setDataDictionary", { newDataDictionary: JSON.parse(data), storeColumns: getters.getColumnNames });
+
+        // 2. Save the filename of the user-provided data dictionary
+        commit("setDataDictionaryFilename", filename);
```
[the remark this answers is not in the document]
jarmoza: Yup! Right from the file input event in `file-selector`. We just haven't used it in this iteration as of yet.

store/index.js
```diff
@@ -798,6 +803,12 @@ export const mutations = {
         p_state.dataDictionary = Object.assign({}, p_state.dataDictionary);
     },
 
+    setDataDictionaryFilename(p_state, p_filename) {
```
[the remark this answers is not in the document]
jarmoza: Added.

cypress/e2e/page/download-pagetests.cy.js
jarmoza: It's worth it to spend a few minutes checking this out. The reason I'm not using Cypress' `readFile` here to check for a file in the downloads path is because our filenames are dynamic in that they contain a timestamp.

However we should be able to use `expect(x).to.contain(first_part_of_filename]`. The possible point of failure there for the test is if the same data dictionary is annotated and downloaded, it would contain the same first part of the filename sans timestamp.

jarmoza: I have found a way to ensure the download folders is clear for tests such as this so the latter test failure possibility has been eliminated.

See: https://codenboxautomationlab.com/how-to-delete-previous-downloaded-files-in-cypress/

cypress/e2e/page/download-pagetests.cy.js
jarmoza: Added this comment

> NOTE: This runs the NodeJS command 'fs.readdirSync' to poll the 'annotation_tool/cypress/downloads folder' as to its current contents

cypress/support/commands.js
```diff
@@ -154,6 +154,53 @@ Cypress.Commands.add("loadAppState", (p_pageName, p_dataset, p_pageData) => {
     } else if ( "download" == p_pageName ) {
 
         // Load state for download page
+
+        // 1. Link all given category column pairs and initialize
```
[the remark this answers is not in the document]
jarmoza: I agree that this function needs to be revisited and perhaps more tightly coupled to the tests as suggested. I've created an issue for this [here](https://github.com/neurobagel/annotation_tool/issues/434). However, for now I am going to keep this code as is.

cypress/unit/store-getter-getJsonOutput.cy.js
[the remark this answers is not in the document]
jarmoza: This may be beyond an aesthetic issue actually. Something I've noticed is that debugging the getters if they are plugged in via the import from `index.js` is not available via console-style debugging. And in fact, it seems that `beforeEach` is working in a different or at least at a different time to properly setup the variables for each test. I've noticed there's a difference between assigning getters to `store.state.getters` in `beforeEach` vs. inside of a test. This is something that should be documented in our Cypress wiki as well.

cypress/e2e/page/download-pagetests.cy.js
jarmoza: I've added a comment to explain it now and there also will be further documentation of this in https://github.com/neurobagel/documentation/issues/29

cypress/e2e/page/download-pagetests.cy.js
jarmoza: After talking with @surchs, instead of the more risky file removal solution I have opted to check to see that the last file added to the `cypress/downloads` folder contains the proper file prefix as saved in `dataDictionary.filename`. Since there were no other tests using the `getVuexStoreValue` method, I have also altered it to look directly at `store.state` instead of trying to look for a getter (an old store paradigm we were using at the time of the method's creation.)

PR #440, opened 2023-06-02
Updated cd.yaml

rmanaem:

- Changed `main` branch to `master` on push event in cd.yml


jarmoza (approved):
🧑‍🍳  @rmanaem 

Unless this becomes an issue yet again in the future with some other tooling, or someone has a strong opinion on the matter, we can just leave the branch name as `master`.

Issue #471, opened 2023-06-21
[BUG] Annotation Tool broken on categorization page on neurobagel.org


jarmoza:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Expected Behavior

Users should be able to categorize columns of their uploaded participants.tsv and proceed from the categorization page to the annotation page

### Current Behavior

The version of the annotation tool currently on `annotate.neurobagel.org` is broken once reaching the categorization page.

Coloring of the category select table and column linking table is absent and the column linking table's rows are no longer clickable.

This blocks proceeding to the annotation page.

The tool functions locally via `npm run dev` and all e2e/component tests are green.

### Error message

![Screen Shot 2023-06-21 at 11 43 21 AM](https://github.com/neurobagel/annotation_tool/assets/4140162/9fccfdcd-651b-4deb-9762-31f82a0f07ba)

### How to reproduce

Go to [annotate.neurobagel.org](annotate.neurobagel.org) and upload a participants.tsv (and optionally, a participants.json). Proceed to the categorization page.

### Anything else?

Fixed in by `bootstrap`/`bootstrap-vue` version rollback in commit https://github.com/neurobagel/annotation_tool/commit/7c71f42c49a51fc0878ef60d8f63fd94fc7daf30

jarmoza:
My initial suspicion was that this had something to do with the static site generation and deployment onto github pages.

However, after testing a few changes with `cd.yaml` (continuous deployment) and `nuxt.config.js` re: https://v2.nuxt.com/deployments/github-pages/, I took the educated guess that this might instead be related to our recent version bumps as pointed out by dependabot – the most directly related being `bootstrap` and `bootstrap-vue`.

After having reverted both to the previous versions, `4.6.1.` and `2.21.2`, respectively, this has now re-enabled the components on the categorization (and presumably on the annotation page as well would we have been able to access it).

jarmoza:
Further discussion on the incompatibility these package(s) and GitHub pages deployment will continue here: https://github.com/neurobagel/annotation_tool/issues/472

surchs:
> The tool functions locally via npm run dev and all e2e/component tests are green.

@jarmoza what happens locally when you run `npm run generate` and the test out the static page locally?

Could you also link the PR (or commit?) where you fixed this, please.

[...]

PR #482, opened 2023-07-05
[MAINT] Updating to bootrap-vue 2.23.1

jarmoza:
Closers #471 and addresses #463.

Re-bumping up `bootstrap-vue` to v2.23.1. This is was a followup to see why the categorization page was breaking in the deployed, static build of the app.

As it turns out, `bootstrap-vue` does not (fully) support `bootstrap` v5. (See [https://bootstrap-vue.org/docs](https://bootstrap-vue.org/docs).

## Checklist

- [ ] PR has an interpretable title with a prefix (`[ENH]`, `[BUG]`, `[DOC]`, `[INFRA]`, `[MAINT]`)
- [ ] PR links to Github issue with mention `Closes #XXXX`
- [ ] Tests pass
- [ ] Code is properly formatted

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "dependencies": {
         "@nuxtjs/axios": "^5.13.6",
         "bootstrap": "^4.6.1",
-        "bootstrap-vue": "^2.21.2",
+        "bootstrap-vue": "^2.23.1",
         "core-js": "^3.19.3",
         "file-saver": "^2.0.5",
         "fs": "^0.0.1-security",
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   "dependencies": {
     "@nuxtjs/axios": "^5.13.6",
     "bootstrap": "^4.6.1",
-    "bootstrap-vue": "^2.21.2",
+    "bootstrap-vue": "^2.23.1",
     "core-js": "^3.19.3",
     "file-saver": "^2.0.5",
     "fs": "^0.0.1-security",


surchs (changes requested):
Thanks @jarmoza!

If this is the PR to close #471, then as I mentioned there we should add at least one test that would have failed before we reverted the Bootstrap major version bump. 

If this PR is just for the version bump, I think it'd be better to close this and have dependabot open one for us directly.

surchs (approved):
Looks good @jarmoza, thanks for addressing the tests. Good to go :cook:

cypress/component/column-linking-table.cy.js
surchs, line 1: Hey @jarmoza. I think if we want to test `cypress/component/column-linking-table.cy.js` here, we should just assert that the style we feed in gets applied. In other words: 

- when I click a row, does the right event get emitted
- when I feed the component a prop, does it get used correctly

are both good component tests. But

- when I click a row, and the right event fires, and I mock a specific mutation that should exist in the store, and then feed that back to the component, it should get applied as I expect

Is not a good component test. I agree that the "is styling applied" question hasn't been tested and we should. But that should be possible by just feeding a specific style and then checking that the style is applied.

The interactive path (click -> emit -> mutate -> propagate back to component -> show) is already handled by the e2e test you wrote.

surchs: > In the case of the column linking table this happens indirectly through reactivity with changes on the columnToCategoryMap.

Yes, this is my point exactly. So let's test "If I change `columnToCategoryMap` and give it to the component, does styling change" **and not** "Can I change `columnToCategoryMap` during the test from inside the prop, then mock the store functionality that creates the reactivity at runtime, and THEN check if styling changes" - that's the store's problem, not ours in this context.

E.g. this will work: 

```js
it("When row is linked, styling is applied", () => {

        cy.mount(ColumnLinkingTable, {

            mocks: {

                $store: store
            },

            computed: Object.assign(store.getters, {
                columnToCategoryMap: () => {

                    return {
                        "participant_id": null,
                        "age": "Age",
                        "sex": null
                    };
                }}),

            plugins: ["bootstrap-vue"],

            propsData: props
        });

        cy.get("[data-cy='column-linking-table-table']").contains("age").parent()
        .should("have.class", "category-style-2")
    });
```

jarmoza: @surchs I get what you're saying about mocking the mutation `alterColumnCategoryMapping` as maybe iffy, but I don't totally agree in this case.

The test you've presented above will check to see if styling occurs properly upon component mount given a mocked Vuex state (re: `columnToCategoryMap`) – which is good but incomplete.

We do actually want to see `styleTableRow` fire when `columnToCategoryMap` is altered post-mount as well, and that would fall under the responsibility of the component. Calls to `alterColumnCategoryMapping` to do so are initiated via `applyCategory`.

Unless you are saying that you feel that `applyCategory` is tested enough via the categorization page e2e test? (The current implementation was my workaround for emulating the behavior of `applyCategory`, both for the reactivity issue I was experiencing and for the sake of time this week.)

jarmoza: @surchs  Addendum to above

>The interactive path (click -> emit -> mutate -> propagate back to component -> show) is already handled by the e2e test you wrote.

So I think you _are_ saying that `applyCategory` is tested enough via the categorization page test?

One further comment then. Should we specify somewhere in our Cypress testing documentation that it is allowable for UI test coverage to be split across component and e2e tests? That's my current takeaway regarding the component test for the column linking table.

PR #483, opened 2023-07-05
Linter fixes for new eslint-plugin-cypress version

jarmoza:
Fixes #468. Since that PR is just merging into master, the order would be to review/merge these changes and then to merge the `eslint-plugin-cypress` afterwards.

Linter fixes include unchaining subsequent actions after clicking on an item in the interface (via an additional statement using `cy.get` on the same html object), and a fix to remove extraneous chaining click call for selecting the data table and data dictionary files.

## Checklist

- [ ] PR has an interpretable title with a prefix (`[ENH]`, `[BUG]`, `[DOC]`, `[INFRA]`, `[MAINT]`)
- [ ] PR links to Github issue with mention `Closes #XXXX`
- [ ] Tests pass
- [ ] Code is properly formatted

### DIFF
--- a/cypress/e2e/page/annotation-pagetests.cy.js
+++ b/cypress/e2e/page/annotation-pagetests.cy.js
@@ -84,7 +84,8 @@ describe("tests on annotation page ui with programmatic state loading and store 
 
                     // B. Select the 'float' transformation heuristic
                     // :data-cy="'selectTransform_' + columnName"
-                    cy.get("[data-cy='selectTransform_age']").click().type("float{enter}");
+                    cy.get("[data-cy='selectTransform_age']").click();
+                    cy.get("[data-cy='selectTransform_age']").type("float{enter}");
 
                     // 5. Assert annotation nav and next page button are enabled
                     cy.assertNextPageAccess("download", true);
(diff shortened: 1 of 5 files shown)


jarmoza:
@rmanaem Starting this off in discussion. Just take a look and I think we can quickly advance it to implement done/review active.

cypress/e2e/page/annotation-pagetests.cy.js
```diff
@@ -84,7 +84,8 @@ describe("tests on annotation page ui with programmatic state loading and store
 
                     // B. Select the 'float' transformation heuristic
                     // :data-cy="'selectTransform_' + columnName"
-                    cy.get("[data-cy='selectTransform_age']").click().type("float{enter}");
+                    cy.get("[data-cy='selectTransform_age']").click();
```
rmanaem, line 87: I think in case of typing action with `v-select`, there is no need to preface it with a `click` action.

rmanaem (approved):
Thanks for the PR @jarmoza! I left a comment but other than that :cook:
Don't forget to add a prefix to the PR title.

cypress/e2e/page/annotation-pagetests.cy.js
jarmoza: Looks green, @rmanaem. Committing the change and will merge.

Issue #545, opened 2023-09-14
Decouple routing logic


rmanaem:
At the moment the routing is being handled implicitly through the `to` attribute of `b-button` and `b-nav-item` in `next-page.vue` and `tool-navbar.vue` respectively. This has led to some confusion and coupling in `next-page.vue` (see #540).
We'd like to separate the logic from a specific element in aforementioned components to allow for isolated testing, debugging, etc.
We could:
- Implement routing in the store e.g., in `setCurrentPage` mutation.
- Implement a separate module that is used to handle routing

PR #546, opened 2023-09-14
[MNT] Removed the test case for checking next page's url

### LINKED ISSUE #540: Rework `next-page` component test
Currently, the last test case in the `next-page.cy.js` intercepts a `GET` request to assert over the url when the `next-page` button is clicked:
https://github.com/neurobagel/annotation_tool/blob/cd4a0a4b911a30c66954e1b985b3315939b095b1/cypress/component/next-page.cy.js#L144-L174

However, since `next-page.cy,js` is a component test, it's better to test the component itself (i.e., it's ability to display elements, fire events, ...) as opposed to how it interacts with the rest of the app. 
So this test case needs to be reworked to check whether the responsible mutation is called when the `next-page` button is clicked.

*This came to light when bumping cypress lead to the test failing in #536*

rmanaem:
Closes #540 

- Removed the test case for checking next page's url, see [this comment](https://github.com/neurobagel/annotation_tool/issues/540#issuecomment-1719907990) for more details
- Includes changes from #536 
- Set the node version of cypress component test to 18 since the `annot-explanation.cy.js` was failing on default node version

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/cypress/component/annot-explanation.cy.js
+++ b/cypress/component/annot-explanation.cy.js
@@ -61,6 +61,8 @@ describe("explanation", () => {
         cy.get(".btn").click();
 
         // Assert
+        // The timeout statement is set here as this test case often takes longer and fails on the CI
+        // while it passes locally. This is a workaround to make the test pass on the CI.
         cy.get(".card-body").should("be.visible",  { timeout: 8000 });
 
         // Act
--- a/cypress/component/annot-explanation.cy.js
+++ b/cypress/component/annot-explanation.cy.js
@@ -61,7 +61,7 @@ describe("explanation", () => {
         cy.get(".btn").click();
 
         // Assert
-        cy.get(".card-body").should("be.visible");
+        cy.get(".card-body").should("be.visible",  { timeout: 8000 });
 
         // Act
         cy.get(".btn").click();
--- a/cypress/component/annot-explanation.cy.js
+++ b/cypress/component/annot-explanation.cy.js
@@ -61,9 +61,7 @@ describe("explanation", () => {
         cy.get(".btn").click();
 
         // Assert
-        // The timeout statement is set here as this test case often takes longer and fails on the CI
-        // while it passes locally. This is a workaround to make the test pass on the CI.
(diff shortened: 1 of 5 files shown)


cypress/component/annot-explanation.cy.js
```diff
@@ -61,7 +61,7 @@ describe("explanation", () => {
         cy.get(".btn").click();
 
         // Assert
-        cy.get(".card-body").should("be.visible");
+        cy.get(".card-body").should("be.visible",  { timeout: 8000 });
```
surchs, line 64: Why is this part of the PR?

package-lock.json
surchs, line 1: seems unrelated

package.json
surchs, line 1: seems unrelated

surchs (commented):
thanks @rmanaem! Agree with removing the test until https://github.com/neurobagel/annotation_tool/issues/545 is addressed. 

There seem to be some unrelated changes in the PR, could you check again that these make sense and are required. In general let's keep the changes in one PR focused to the issue so we can roll them back more easily if something breaks, and can reason better over what might be causing a new bug/breaking.

cypress/component/annot-explanation.cy.js
rmanaem: This was inconsistent across different ci runs. For example, It failed [here](https://github.com/neurobagel/annotation_tool/actions/runs/6189686027/job/16805322365) when I was resolving the conflict.
It always passed locally so I checked the logs and it seemed to be timing out sometimes.

package.json
rmanaem: See #536

package-lock.json
rmanaem: See #536

cypress/component/annot-explanation.cy.js
surchs: ok, that's good to know. I'd say: add a short comment above this line to say that and then good to go

surchs (approved):
thanks, let's add a brief explainer comment for the custom timeout and then :cook:

PR #555, opened 2023-09-25
[MNT] Refactored `annot-explanation` component

### LINKED ISSUE #550: Check if `annot-explanation.cy.js` is still flaky
Since bumping cypress to 13.0.0, the `annot-explanation.cy.js` has been acting flaky i.e., passing and failing randomly (see #546).
We pinned the node version in the cypress component test action and it seemed to resolve the issue. Let's make sure that's the case.
Resources:
[Cypress docs](https://docs.cypress.io/guides/cloud/flaky-test-management)

- [ ] understand or at least get a good intuition on what is causing this test to fail. Although the test is not important, understanding how cypress' changes are impacting previously running tests is. So we may have to spend more time than the tiny test itself would warrant to figure this out
- [ ] once we know what's wrong, see if we can refactor the test / remove it. as far as I can tell, currently the test is almost testing the bootstrap-accordion component itself

rmanaem:
Closes #550 

- Removed `b-collapse` component from `annot-explanation` component
- Updated the `annot-explanation` component test

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -26,9 +26,6 @@ updates:
         update-types: ["version-update:semver-major"]
       - dependency-name: 'eslint-plugin-prettier'
         update-types: ["version-update:semver-major"]
-      - dependency-name: 'cypress-io/github-action'
-        update-types: ["version-update:semver-major"]
-
     labels:
       - "_bot"
       - "maint:dependency"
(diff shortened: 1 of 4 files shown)


surchs (approved):
Looks good @rmanaem :tada:. I agree that this is the least bad solution given that we cannot figure out why this specific element is suddendly making our tests flaky. 

This does the job well, and, importantly, doesn't seem to break!

:cook:

PR #586, opened 2023-10-26
[ENH] Implemented annoatation workflow for healthy controls

### LINKED ISSUE #553: Allow annotation of healthy controls
Right now there is no way for a user to indicate that a specific value in the data-table means that the corresponding subject is a healthy control. Further, there is no clear design for a workflow that would let a user provide this annotation. E.g. if there is a column "healthy control" with values "True" and "False", what should they do. Or if the information is in a diagnostic column

We decided to use the following workflow

- any column that has info on "isControl" should get classified as about "Diagnosis"
- on the Annotation page, we add a "isControl" button
- the user should click this button to signal that a specific value in the "Diagnosis" column indicates that the subject is a control
- if the column only contains info on isControl (e.g. dummy coding, True/False), then the remaining values should be classified as missing

- [x] Design a generic workflow to annotate "healthy control", reusing existing workflows as much as possible
- [ ] Implement the workflow by adding the required UI components and logic
- [ ] provide tests for these cases:
  - healthy-control status is part of a generic "Diagnosis" column
  - healthy-control status is dummy-coded, e.g. as a "is_healthy_control" column
  - there is no information on healthy control / the annotation is not made
- [ ] make sure the is_control status successfully makes it into the output .json

rmanaem:
Closes #553 

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:

### DIFF
--- a/components/annot-categorical.vue
+++ b/components/annot-categorical.vue
@@ -40,6 +40,14 @@
                             {{ uiText.missingValueButton }}
                         </b-button>
                     </template>
+                    <template #cell(isControl)="row" v-if="activeCategory == 'Diagnosis'">
+                        <b-button
+                            :data-cy="'isControlButton_' + row.index"
+                            variant="info"
+                            @click="selectOptionAndUpdate('ncit:C94342', row.item['columnName'], row.item['rawValue'])">
+                            {{ uiText.isControlButton }}
+                        </b-button>
+                    </template>
 
                 </b-table>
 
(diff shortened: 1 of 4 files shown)


components/annot-categorical.vue
```diff
@@ -40,6 +40,14 @@
                             {{ uiText.missingValueButton }}
                         </b-button>
                     </template>
+                    <template #cell(isControl)="row" v-if="activeCategory == 'Diagnosis'">
```
surchs, line 43:
```suggestion
                    <template #cell(isControl)="row" v-if="activeCategory === 'Diagnosis'">
```

surchs (approved):
:cook: with the one comment!

PR #623, opened 2023-11-10
[FIX] read value description from user dictionary correctly

### LINKED ISSUE #616: Descriptions for discrete values are not shown in the annotation workflow
At the moment, any "Levels" descriptions we get from the BIDS data dictionary (i.e. for discrete columns) are just ignored in the annotation views:

image

but see

image

This is (hopefully) and easy fix, but to make sure it doesn't come back we need both:

- [ ] Show "Levels" descriptions for values on the appropriate discrete annotation workflows (i.e. "Sex" and "Diagnosis"?)
- [ ] Write a test (component ?) that makes sure this actually works (and stays that way)

surchs:
- Closes #616 

A couple of things I noticed while working on this:
1. If we mock store data incorrectly, i.e. not how it actually looks at runtime, then the tests pass but the app is still broken
2. It's really not very easy to get a quick sense of where information is kept in the store or what it would look like at runtime. The dev tools help, but it's still tricky to understand how a value ended up looking the way it does
3. Our e2e should help catch this but they are really hard to edit / change, so it's often easier to test manually - that's not good

- fix store getter for value descriptions
- fix mocked data in unit test for getter

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:


store/index.js
```diff
@@ -644,9 +644,9 @@ export const getters = {
     getValueDescription: (p_state) => (p_columnName, p_value) => {
         // Returns the description of a value in a column, if that description exists
         // Otherwise it returns an empty string
-        const description = p_state.dataDictionary.annotated[p_columnName].levels?.[p_value]?.description;
+        const description = p_state.dataDictionary.annotated[p_columnName].Levels?.[p_value];
         if ( typeof description  === "undefined" ) {
-            return "";
+            return "no description available";
```
rmanaem, line 649: Nice! explicit > implicit

rmanaem (approved):
:cook:

PR #627, opened 2023-11-12
[ENH] Improved column highlighting in assessment tool workflow

### LINKED ISSUE #614: Improve column highlighting in assessment tool workflow
We use colors to give the user visual feedback on what tool is selected and what columns have already been mapped to it. 

image

This should be improved in the following ways

- [x] find a color that is less :boom: 
- [x] tools that are active should be colored, inactive tools should be white (as now)
- [x] columns that are not mapped to **any** tool should be white (as now)
- [x] columns that are mapped to a tool should not be white
- [x] if the column is mapped to the currently **active** (i.e. selected) tool, then the column should be brightly colored, otherwise more muted 

Essentially: make it more like the column-categorization workflow above, but with only one color

rmanaem:
Closes #614 

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:


components/category-toolgroup.vue
```diff
@@ -96,31 +96,29 @@
 
             styleTableRow(p_row) {
                 if (p_row.identifier === this.selectedTool.identifier) {
-                    return "selected-tool";
+                    return "category-style-5";
```
surchs, line 99: I think this works, but we should make an issue to refactor the whole color handling in the store. Because now we're coupled to a specific style name here and in all other places we directly have to read the mapping of categories to styles from the store. 

Would you mind making an issue to go over this whole section here https://github.com/neurobagel/annotation_tool/blob/1161c37e912cfd66fde376e1a6c8755c87a223bd/store/index.js#L62-L91 and discussing how to make it easier to use?

cypress/component/category-toolgroup.cy.js
```diff
@@ -220,15 +220,26 @@ describe("Tool Group component", () => {
...
             cy.get("[data-cy='assessment-tool-table']")
             // Selecting the tool my column is mapped to should change my columns styling
             .find("tr:contains('UPDRSIII')").click();
-            // assert that element has different color after
-            cy.get("[data-cy='assessment-column-table']").find("tr:contains('column3')").should("not.have.css", "background-color", InitialBackgroundColor);
+            // assert that element has the same background color after
+            cy.get("[data-cy='assessment-column-table']").find("tr:contains('column3')").should("have.css", "background-color", InitialBackgroundColor);
+        });
+
+        // Starting out with a different tool selected and then checking the opacity
```
surchs, line 234: nice :tada:

surchs (approved):
nice :tada:! Tried it out (thanks netlify), looks good. I left a comment for us to revisit.

Otherwise :cook:

components/category-toolgroup.vue
rmanaem: Done #631

PR #650, opened 2023-11-29
[FIX] allow users to use the browser-back button

### LINKED ISSUE #491: [ENH] Add component test for tool-navbar
### Is there an existing issue for this?

- [X] I have searched the existing issues

### New feature

_No response_

### Unclear documentation

We currently don't have any tests for the `tool-navbar` component. Look to the query tool's tests for its navbar for comparison.

### LINKED ISSUE #619: Add warning not to use browser back button
There should be a warning on the bottom of the page not to click the "back" button on the browser (which seems to cause some elements to persist weirdly) but use the navigator at the top of the app instead.

surchs:
- Closes #619 
- Closes #491

A lot of the stuff in the store around route handling and pages is already taken care of by the router. This PR essentially removes all of this stuff and replaces it with functions from the router

One big issue with the old system was that it **only** looked to the store to find out what page it was currently on. so if I use the browser back button this is not reflected in the store, but the route changes. So I could end up e.g. on the home page with a state that thinks I'm on the annotation page. 

So we can make this both easier and have it not break anymore

- add pageOrder in the store (tricky to do in the router)
- renamed index.vue to home.vue -> so the route name works out
- added test for nav-bar

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:


nuxt.config.js
```diff
@@ -1,4 +1,19 @@
...
+			routes.push({
+			path: '/',
+			name: 'home',
+			components: {
+			// eslint-disable-next-line no-undef
+			default: resolve(__dirname, 'pages/home.vue')
+			}
+		});
+		}
+	},
```
rmanaem, line 16: Interesting, didn't know we could do this. Did u find this in their docs?

layouts/default.vue
```diff
@@ -9,27 +9,17 @@
         <Nuxt />
 
         <!-- Next page button -->
-        <next-page v-if="getNextPage()!==''" />
```
rmanaem, line 12: I believe we still need a condition to prevent rendering `next-page` component when we get to the download page. Currently you can see a small gray button displayed.

```suggestion
        <next-page v-if="$route.name !== 'download'"/>
```

rmanaem (changes requested):
Nicely done @surchs! Handling of routes is elegant and easy to understand.
Aside from the comments, I have two 🍒 
- I think we can remove `getNextPage` and `setCurrentPage` from the store.
- I'd suggest adding a test for using the browser back button to one of the e2e tests.

nuxt.config.js
surchs: it went a little like this:
step 1) ask chatgpt, step 2) confirm: https://v2.nuxt.com/docs/configuration-glossary/configuration-router/#extendroutes

rmanaem (changes requested):
I think you forgot to remove `getNextPage` and `setCurrentPage` from the store and add a test for the browser back button.

surchs:
@rmanaem can you confirm where you still see `getNextPage` and `setCurrentPage` in the store? I saw them in the test and removed them there.

rmanaem:
> can you confirm where you still see getNextPage and setCurrentPage in the store?

https://github.com/neurobagel/annotation_tool/blob/86c0abff4940827bb86fac968ab0671d82c82f1a/store/index.js#L497-L518
https://github.com/neurobagel/annotation_tool/blob/86c0abff4940827bb86fac968ab0671d82c82f1a/store/index.js#L839-L842

rmanaem (approved):
Thanks @surchs! 🧑‍🍳

PR #655, opened 2023-12-09
[ENH] Implemented styling for table rows of `annot-single-tool` component

### LINKED ISSUE #620: Cleaner visual interface for assessment tool missing value annotation
Since all the unique for a given column are presented to be annotated as missing on the same page, it would be helpful if all column values were visually grouped in some way (with a gap between columns, color-coded, etc.) to make it a bit easier to keep track of what column you are annotating.

image

TODO
- [ ] Figure out what visual elements are practical to highlight the visual separation between rows that belong to the same column
- [ ] propose the visual element (check off with someone else)
- [ ] implement this visual element (assuming the table will not be sortable)
- [ ] write a test for it

rmanaem:
- Closes #620 

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:

### DIFF
--- a/components/annot-single-tool.vue
+++ b/components/annot-single-tool.vue
@@ -58,7 +58,9 @@
                 const columns = this.uniqueColumnValues.map(item => item.column);
                 const uniqueColumns = [...new Set(columns)];
                 const colIndex = uniqueColumns.indexOf(row.column);
+                // Apply special styling if we are in a row that belongs to an even column
                 if (colIndex % 2 === 0) {
+                    // Apply "stripe" styling if we are in an even row
                     if (this.uniqueColumnValues.indexOf(row) % 2 === 0) {
                         styleClass.push('column1-color1');
                     }
--- a/components/annot-single-tool.vue
+++ b/components/annot-single-tool.vue
@@ -2,10 +2,10 @@
 
     <div>
         <b-table
-            striped
             :data-cy="('tool-annotation-for-' + name)"
             :fields="toolFields"
-            :items="uniqueColumnValues">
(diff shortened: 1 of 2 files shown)


components/annot-single-tool.vue
```diff
@@ -52,8 +52,44 @@
...
                 this.$emit("declareMissing", {column, value});
+            },
+            styleTableRow(row) {
+                const styleClass = [];
+                const columns = this.uniqueColumnValues.map(item => item.column);
+                const uniqueColumns = [...new Set(columns)];
+                const colIndex = uniqueColumns.indexOf(row.column);
+                if (colIndex % 2 === 0) {
+                    if (this.uniqueColumnValues.indexOf(row) % 2 === 0) {
+                        styleClass.push('column1-color1');
```
surchs, line 63: this is clear or at least I don't know how to make it much clearer, but I still need to read a couple times to figure out row vs column index. So let's add a comment

```suggestion
                // Apply special styling if we are in a row that belongs to an even column
                if (colIndex % 2 === 0) {
                    // Apply "stripe" styling if we are in an even row
                    if (this.uniqueColumnValues.indexOf(row) % 2 === 0) {
                        styleClass.push('column1-color1');
```

surchs (approved):
very nice PR, thanks @rmanaem!

I left one suggestion for clarity there. One other thing I noticed is that if we ever decide to enable table sorting, we probably have to rethink this. We'll probably figure this out then, but you could also add a TODO comment as a reminder.

:cook:

PR #671, opened 2024-01-09
[ENH] Add validation to data dictionary before download

### LINKED ISSUE #637: Warn user about incomplete annotation before downloading
At the moment, a user can make an incomplete annotation (i.e. leave out some required steps to make valid neurobagel .json) and our annotation tool will not say anything / emit any warning. 

An example:

If
- I label a column as "about" "Age"
- I then never pick any "Transformation" for the "Age" columns (e.g. because I forget)

Then
- the annotation tool will happily turn the "next" button green and let me proceed to the download page
- and will let me download a .json that is not actually valid under our own schema (because it is missing the "Transformation" key). See also #597

I think we originally made that change because we thought a user might want to stop half-way through an annotation and pick it up later - and we shouldn't prevent them from downloading what they have until then. However, our tool isn't yet able to load our own annotations again. And even if we want to support this, we should definitely give a warning.

TODO:

- [x] validate the JSON we're about to give back to the user against our data dictionary json schema
- [x] Emit a warning on the download page or before if a user proceeds to download a .json file that isn't valid / user hasn't completed all annotation steps
- [ ] propose a way to keep up to date on the data dictionary schema

surchs:
- Closes #637 

bad route:
image

happy path:
image

-
-

## Checklist

- [ ] Tests pass
- [ ] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:


alyssadai:
One suggestion: maybe we can add to the warning the fact that if you exist the annotation tool after having downloaded an incomplete annotation, you cannot reload the partial annotation back in

cypress/e2e/app/validate-output-example-synthetic-e2etest.cy.js
```diff
@@ -110,5 +119,27 @@ describe("to annotate an assessment ", () => {
               });
         });
     });
+    it.only("shows me an error if I didn't finish the annotation, creating invalid output", () => {
```
rmanaem, line 122: Left over from debugging?

pages/download.vue
```diff
@@ -95,10 +129,30 @@
             defaultOutputFilename() {
 
                 return `${this.dataDictionaryFilenameNoExtension}_annotated_${Date.now()}.json`;
+            },
+
+            outputIsValid() {
+                // Perform JSON schema validation and update validationError
+                const ajv = new Ajv();
+                const isValid = ajv.validate(jsonSchema, this.getJsonOutput);
+                console.log("outputIsValid", isValid);
```
rmanaem, line 138: Debugging left over

rmanaem (approved):
Two small things left over from debugging other than that :cook:

PR #675, opened 2024-01-15
[ENH] Add session ID as a new category for annotation

### LINKED ISSUE #663: can annotate "Session ID" as phenotypic session (but can be empty if only for the purpose of listing the imaging session)
Now that data dictionaries can have annotations on `session_id`: 
https://github.com/neurobagel/neurobagel_examples/blob/519b076bc991fd4549632986874d52f9ca821acc/data-upload/example_synthetic.json#L12-L20

the annotation tool needs to be able to label a column as "about" sessionID

TODO:
- [x] New category `Session ID`
- [x] session_id is not required and does not have a workflow
- [x] when selected, `session_id` appears in the output (and is valid as per the schema)

surchs:
- Closes #663 

- add `Session ID` as a category
- add new color for session id
- make sure session ID does not count as an annotation and does not get its own annotation workflow (like Subject ID)
- add a simple e2e test to make sure of this

Note: this should probably have more extensive tests, I went for a basic test because we may soon refactor the app. But could add appropriate tests as well.

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:

### DIFF
--- a/cypress/e2e/page/categorization-pagetests.cy.js
+++ b/cypress/e2e/page/categorization-pagetests.cy.js
@@ -128,6 +128,40 @@ describe("Tests on categorization page ui via programmatic state loading and sto
                             .should("not.have.css", "background-color", p_oldColor);
                     });
             });
+
+            it.only("doesn't enable the next button until something other than Subject ID or Session ID is clicked", () => {
+                // Note: skipping the fancy test dataset handling since we only have one test dataset
+                cy.assertNextPageAccess("annotation", false);
+
+                cy.get("[data-cy='categorization-table']")
+                    .contains("Subject ID")
+                    .click();
+                cy.get("[data-cy='column-linking-table'] tbody > tr > td")
+                    .contains("participant_id")
+                    .click();
+
+                // Still expect the next page to be inaccessible
+                cy.assertNextPageAccess("annotation", false);
+
+                cy.get("[data-cy='categorization-table']")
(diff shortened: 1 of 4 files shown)


surchs:
> good idea to update the test containing the output validation using schema to ensure the output is valid after the session ID is annotated

Yeah, I was thinking about that. The main thing is that our current example files don't have a session_id column and I didn't want to change the example / introduce a new one. I could just have it label an incorrect column as session_id and see what happens. Or do you have a better idea?

cypress/e2e/page/categorization-pagetests.cy.js
```diff
@@ -128,6 +128,40 @@ describe("Tests on categorization page ui via programmatic state loading and sto
...
+
+                // Still expect the next page to be inaccessible
+                cy.assertNextPageAccess("annotation", false);
+
+                cy.get("[data-cy='categorization-table']")
+                    .contains("Session ID")
+                    .click();
+                cy.get("[data-cy='column-linking-table'] tbody > tr > td")
+                    .contains("age") // TODO: eventually find a good column for this stuff
+                    .click();
```
surchs, line 151: hey @rmanaem, I just realized that the "just assign session_id to something" step is already happening here. Maybe you missed it. Unless you have a different suggestions, I'd merge as is? sound good?

PR #715, opened 2024-04-10
[FIX] Remove existing annotations from data dictionaries

### LINKED ISSUE #677: If a user provides an annotated dictionary, the app ignores it but keeps the annotations
We do not yet support uploading an already annotated data dictionary (see #601). However, if a user regardless uploads an already annotated data dictionary the app will happily take it. This leads to some very confusing states:

- if I provide a completely annotated data dictionary :shrug: then the visual state of the app will tell me nothing is annotated yet, but my download will include all my existing annotations
- if I provide an invalid data dictionary then again, the UI looks empty, but when I come to the download page and my data dictionary is validated, the validation will fail. Unless I guess what's going on, I cannot fix that

So in short: the state of the data dictionary in the app and the visual UI do not correspond to each other, and that's quite bad. 
We have two options:
1. Teach the app to load state from a data dictionary (#601)
2. Put a big red warning on the first page that says: you cannot load a data dictionary with neurobagel annotations

I lean towards 2 because we will likely refactor this soon

TODO:
- [x] when existing data dictionary is loaded, strip away / rest the "Annotations" part
- [x] big warning on Home page near the "data dictionary" upload button to say "cannot reuse annotations"
- [x] test that existing annotations do not make it through the tool, e.g. no unintentional annotations

surchs:
- Closes #677 

- expandable warning button on home page to say "can't reuse"
- strip "Annotations" keys from data dictionary after loading in store
- note: there is no test for this because the `processDataDictionary` store method never had a test written ...

## Checklist

- [x] PR has an interpretable title with a prefix (`[ENH]`, `[FIX]`, `[REF]`, `[TST]`, `[CI]`, `[MNT]`, `[INF]`, `[MODEL]`, `[DOC]`) _(see our [Contributing Guidelines](https://neurobagel.org/contributing/CONTRIBUTING#pull-request-guidelines) for more info)_
- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/pages/home.vue
+++ b/pages/home.vue
@@ -71,6 +71,16 @@
         <b-row>
             <b-col>
                 <h3>{{ uiText.dataDictionaryHeader }}</h3>
+                <b-button data-cy="cannot-reuse-annotations-button" v-b-toggle.collapse-2 variant="warning">Cannot reuse annotations</b-button>
+                <b-collapse id="collapse-2">
+                    <br />
+                    At the moment, the annotation tool is not able to load previously created annotations.
+                    This means thart if you provide a data dictionary here that you created with Neurobagel,
+                    any existing semantic annotations will be removed when the data dictionary is loaded.
+
+                    In practice this means that you cannot resume an aborted previous annotation.
+                    This is a limitation we are going to address: <a href="https://github.com/neurobagel/annotation_tool/issues/601" target="_blank">#601</a>.
+                </b-collapse>
                 <file-selector
                     data-cy="data-dictionary-selector"
                     :content-type="contentTypes.dataDictionary"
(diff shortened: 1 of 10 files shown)


pages/home.vue
```diff
@@ -71,6 +71,16 @@
         <b-row>
             <b-col>
                 <h3>{{ uiText.dataDictionaryHeader }}</h3>
+                <b-button data-cy="cannot-reuse-annotations-button" v-b-toggle.collapse-2 variant="warning">Cannot reuse annotations</b-button>
+                <b-collapse id="collapse-2">
+                    <br />
+                    At the moment, the annotation tool is not able to load previously created annotations.
+                    This means thart if you provide a data dictionary here that you created with Neurobagel,
```
rmanaem, line 78:
```suggestion
                    This means that if you provide a data dictionary here that you created with Neurobagel,
```

rmanaem (approved):
Just a typo fix and then :cook:

PR #832, opened 2024-12-12
[ENH] Implemented `Annotate Another Dataset` button

### LINKED ISSUE #816: Add "Annotate another dataset" button on the download page that refreshes the app
Currently, if I navigate back to the "Home" tab after completing an annotation, it seems that the tool is not fully "refreshed". For example, it seems the assessment tools I selected from the dropdown to annotate the previous dataset would still be selected by default.

rmanaem:
- Closes #816

## Checklist

- [x] PR has an interpretable title with a prefix (`[ENH]`, `[FIX]`, `[REF]`, `[TST]`, `[CI]`, `[MNT]`, `[INF]`, `[MODEL]`, `[DOC]`) _(see our [Contributing Guidelines](https://neurobagel.org/contributing/CONTRIBUTING#pull-request-guidelines) for more info)_
- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/store/index.js
+++ b/store/index.js
@@ -944,8 +944,8 @@ export const mutations = {
     },
 
     resetState(p_state) {
+        const initialState = getDefaultState();
         Object.keys(p_state).forEach((key) => {
-          const initialState = getDefaultState();
           if (Object.prototype.hasOwnProperty.call(initialState, key)) {
             Vue.set(p_state, key, initialState[key]);
           }
--- a/store/index.js
+++ b/store/index.js
@@ -4,7 +4,9 @@ import Vue from "vue";
 import toolTerms from "~/static/toolTerms.json";
 import diagnosisTerms from "~/static/diagnosisTerms.json";
 
-export const state = () => ({
+// The default state of the store
+function getDefaultState() {
+    return {
 
     // Keeps track of number of annotations made in current run;
     // > 0 annotations allows access to the download page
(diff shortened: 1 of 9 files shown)


store/index.js
```diff
@@ -936,5 +941,14 @@ export const mutations = {
 
         // 2. Save the updated annotation in the store
         p_state.annotationCount = count;
-    }
+    },
+
+    resetState(p_state) {
+        Object.keys(p_state).forEach((key) => {
+          const initialState = getDefaultState();
+          if (Object.prototype.hasOwnProperty.call(initialState, key)) {
```
surchs, line 949:
```suggestion
          if (Object.hasOwn(initialState, key)) {
```

see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn and we don't plan on supporting older browsers

We could most likely also just go with `initialState.hasOwnProperty(key)` here because I think your version mainly guards against the possibility of `"hasOwnProperty"` being an actual key in the store - which we don't have.

surchs (approved):
One suggestion, otherwise :cook: 

sidenote: The diff on these chained PRs is a bit tricky to read. It works because they are short, but even happier would be either:
- last PR merges into previous branch it chained off of (assuming that is not merged yet)
- every PR is branched off (and targeting) `main` directly.

store/index.js
rmanaem: Tried it, didn't work.

PR #864, opened 2025-03-05
[ENH] Add fromRanged as age transformation

### LINKED ISSUE #862: Bring back ranged ages
Corresponding change to https://github.com/neurobagel/bagel-cli/issues/433

Bottom line is: we need to bring back the `ranged` format for ages that we removed following #392. 

TODO:

- [x] Add back a "format" option called `ranged`
- [x] have it map to `nb:FromRange` in the output
- [x] add a test for this new format

surchs:
- Closes #862

-
-

## Checklist

- [ ] PR has an interpretable title with a prefix (`[ENH]`, `[FIX]`, `[REF]`, `[TST]`, `[CI]`, `[MNT]`, `[INF]`, `[MODEL]`, `[DOC]`) _(see our [Contributing Guidelines](https://neurobagel.org/contributing/CONTRIBUTING#pull-request-guidelines) for more info)_
- [ ] PR links to GitHub issue with mention `Closes #XXXX`
- [ ] Tests pass
- [ ] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/store/index.js
+++ b/store/index.js
@@ -143,6 +143,12 @@ function getDefaultState() {
 
             Label: "period of time defined according to the ISO8601 standard",
             TermURL: "nb:FromISO8601"
+        },
+
+        range: {
+
+            Label: "Age defined as a range of two integers",
+            TermURL: "nb:FromRanged"
         }
     },
 
@@ -374,6 +380,11 @@ export const getters = {
             case "int":
 
                 convertedValue = parseInt(p_originalValue);
+                break;
+
+            case "range":
+
+                convertedValue = p_originalValue.split("-").reduce((acc, value) => parseInt(acc) + parseInt(value)) / 2;
                 break;
 
             // case "isoyear": {


store/index.js
```diff
@@ -143,6 +143,12 @@ function getDefaultState() {
 
             Label: "period of time defined according to the ISO8601 standard",
             TermURL: "nb:FromISO8601"
+        },
+
+        range: {
+
+            Label: "Age defined as a range of two integers",
+            TermURL: "nb:FromRanged"
```
alyssadai, line 151: Hey @surchs, I wanted to quickly double check how the annotation was defined here to match the corresponding CLI [PR](https://github.com/neurobagel/bagel-cli/pull/437):

- I think we decided in the [issue](https://github.com/neurobagel/legacy_annotation_tool/issues/862#issuecomment-2683465926) to call this `nb:FromRange` to match our other format names!
- Do you think the integer constraint is important? For the CLI at least, numbers on either side of `-` are converted to float before the average is calculated (this is also consistent with the legacy behaviour) so whether the min/max are integers or floats doesn't really matter 🤷

surchs: aha, good point, thanks. I went off the issue description - seems we didn't update to `nb:FromRange` there. 

Yeah, int is probably not very important, we can go with Float. I'll make those changes

rmanaem (approved):
Looks good to go! :cook:

